PR #60, opened 2022-02-04
Fixed BGP regression failure bugs

stalabi1:
##### SUMMARY

- The advertise-list in the OC yang was removed so the advertise_prefix in the sonic_bgp_af module needed to be removed.
- Attributes in the actioner were changed to int so those attributes were changed to int in the sonic_bgp_neighbors module

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

sonic_bgp_af, sonic_bgp_neighbors

##### ADDITIONAL INFORMATION

[regression-2022-02-03-10-45-58.pdf](https://github.com/ansible-collections/dellemc.enterprise_sonic/files/8006609/regression-2022-02-03-10-45-58.pdf)
[bgp_af_regression_report.pdf](https://github.com/ansible-collections/dellemc.enterprise_sonic/files/8007057/bgp_af_regression_report.pdf)

### DIFF
--- a/tests/regression/roles/sonic_bgp_af/defaults/main.yml
+++ b/tests/regression/roles/sonic_bgp_af/defaults/main.yml
@@ -89,11 +89,6 @@ tests:
             - afi: l2vpn
               safi: evpn
               advertise_all_vni: True
-              advertise_prefix:
-                - afi: ipv4
-                  safi: unicast
-                - afi: ipv6
-                  safi: unicast
       - bgp_as: "{{bgp_as_2}}"
         vrf_name: "{{vrf_1}}"
         address_family:
@@ -130,11 +125,6 @@ tests:
                   route_map: rmap_reg2
             - afi: l2vpn
               safi: evpn
-              advertise_prefix:
-                - afi: ipv4
-                  safi: unicast
-                - afi: ipv6
-                  safi: unicast
   - name: test_case_03
     description: Update2 created BGP AF properties
     state: merged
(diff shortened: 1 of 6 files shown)


tests/regression/roles/sonic_bgp_af/defaults/main.yml
```diff
@@ -298,7 +288,3 @@ tests:
         vrf_name: default
         address_family:
           afis:
-  - name: test_case_07
-    description: Delete3 BGP AF properties
-    state: deleted
-    input: []
```
kerry-meyer, line 304: Was this test case failing? (It is common to do a "delete all" at the end of a test. Usually, it doesn't cause a failure.)

kerry-meyer:
Most of my comments from PR 59 went away with the upstream fetch because those diffs were already present in the "main" branch. My only remaining question is regarding removal of the bgp_af "delete all" test case.

All of the other remaining changes look fine to me and I have no questions on them.

tests/regression/roles/sonic_bgp_af/defaults/main.yml
stalabi1: Yes it was failing the after was already undefined

kerry-meyer: We need to do some analysis to figure out the reason for the failure.

stalabi1: Test case 7 was failing due to being undefined since the previous test case already deleted all existing config. I removed config from test case 5 so that the delete all would have something to delete.

kerry-meyer:
The revised change set (with the restored test) looks good.

Approved.

kerry-meyer (approved):

PR #69, opened 2022-04-21
double commit for bgp neighbors password and description attributes

stalabi1:
##### SUMMARY
Double commit for bgp neighbors module password and description attributes

[regression-2022-04-21-12-12-17.html.pdf](https://github.com/ansible-collections/dellemc.enterprise_sonic/files/8534451/regression-2022-04-21-12-12-17.html.pdf)


kerry-meyer (approved):
Approved.

Thanks for doing the double commit.

PR #82, opened 2022-05-19
Added static routes module

stalabi1:
##### SUMMARY

Added static routes module

##### ISSUE TYPE

- New Module Pull Request

##### COMPONENT NAME

sonic_static_routes

##### ADDITIONAL INFORMATION

[static_routes_facts_UT.log](https://github.com/ansible-collections/dellemc.enterprise_sonic/files/8721638/static_routes_facts_UT.log)
[show_running_configuration_static_routes.txt](https://github.com/ansible-collections/dellemc.enterprise_sonic/files/8721646/show_running_configuration_static_routes.txt)
Updated:
[regression-2022-06-17-14-59-13.html.pdf](https://github.com/ansible-collections/dellemc.enterprise_sonic/files/8931517/regression-2022-06-17-14-59-13.html.pdf)
[static_routes_output.log](https://github.com/ansible-collections/dellemc.enterprise_sonic/files/8931519/static_routes_output.log)


plugins/module_utils/network/sonic/config/static_routes/static_routes.py
```diff
@@ -0,0 +1,341 @@
...
+                vrf_name = cmd.get('vrf_name', None)
+                static_list = cmd.get('static_list', [])
+                for static in static_list:
+                    prefix = static.get('prefix', None)
+                    next_hops = static.get('next_hops', [])
+
+                    for cfg in have:
+                        cfg_vrf_name = cfg.get('vrf_name', None)
+                        if vrf_name == cfg_vrf_name:
+                            cfg_static_list = cfg.get('static_list', [])
```
kerry-meyer, line 273: Can the user delete all prefixes for a VRF by specifying a VRF with no prefix list?

If so, handling for that case needs to be added.

plugins/module_utils/network/sonic/config/static_routes/static_routes.py
```diff
@@ -0,0 +1,341 @@
...
+        return idx
+
+    def get_delete_static_routes_requests(self, commands, have, is_delete_all):
+        requests = []
+        if is_delete_all:
+            requests = self.get_delete_all_static_routes_requests(commands)
+        else:
+            for cmd in commands:
+                vrf_name = cmd.get('vrf_name', None)
+                static_list = cmd.get('static_list', [])
```
kerry-meyer, line 265: Can the user delete all prefixes for a VRF by specifying a VRF with no prefix list?

If so, handling for that case needs to be added.

plugins/module_utils/network/sonic/config/static_routes/static_routes.py
```diff
@@ -0,0 +1,341 @@
...
+                                            cfg_next_hops = cfg_static.get('next_hops', [])
+                                            if cfg_next_hops:
+                                                for cfg_next_hop in cfg_next_hops:
+                                                    cfg_index = cfg_next_hop.get('index', {})
+                                                    cfg_idx = self.generate_index(cfg_index)
+                                                    if idx == cfg_idx:
+                                                        cfg_metric = cfg_next_hop.get('metric', None)
+                                                        cfg_track = cfg_next_hop.get('track', None)
+                                                        cfg_tag = cfg_next_hop.get('tag', None)
+                                                        if idx and not metric and not track and not tag:
```
kerry-meyer, line 296: How can "idx" not exist here?

stalabi1: there's no not in front of idx

plugins/module_utils/network/sonic/facts/static_routes/static_routes.py
```diff
@@ -0,0 +1,172 @@
...
+                        next_hop_dict['track'] = track
+                    if tag:
+                        next_hop_dict['tag'] = tag
+                    if next_hop_dict and next_hop_dict not in next_hops_list:
+                        next_hops_list.append(next_hop_dict)
+                if prefix:
+                    static_dict['prefix'] = prefix
+                if next_hops_list:
+                    static_dict['next_hops'] = next_hops_list
+                if static_dict and static_dict not in static_list:
```
kerry-meyer, line 164: Same question here: How could the static_dict already be in the list? (The same question applies for the static_vrf_dict presence in the static_vrf_list, below)

plugins/modules/sonic_static_routes.py
```diff
@@ -0,0 +1,241 @@
...
+        type: list
+        elements: dict
+        description:
+          - A list of 'static_routes' configurations.
+        suboptions:
+          prefix:
+            required: True
+            type: str
+            description:
+              - List of locally configured static routes.
```
kerry-meyer, line 62: This description doesn't look correct.

The Yang description says:

"Destination prefix for the static route, either IPv4 orIPv6."

plugins/modules/sonic_static_routes.py
```diff
@@ -0,0 +1,241 @@
...
+    default: merged
+"""
+EXAMPLES = """
+
+# Using merged
+#
+# Before State:
+# -------------
+#
+# sonic# show running-configuration | grep "ip route"
```
kerry-meyer, line 120: It looks like this line was added by mistake.

plugins/module_utils/network/sonic/config/static_routes/static_routes.py
kerry-meyer: What I meant was that "if idx and not metric ..." could just be "if not metric ..." because idx is known to exist.

plugins/modules/sonic_static_routes.py
stalabi1: I wanted to show that the command returned empty because there is no preexisting config before the merge

plugins/module_utils/network/sonic/facts/static_routes/static_routes.py
```diff
@@ -0,0 +1,172 @@
...
+                static_dict = {}
+                prefix = static.get('prefix', None)
+                next_hops = static.get('next-hops', None)
+                next_hop_list = next_hops.get('next-hop', [])
+                next_hops_list = []
+                for next_hop in next_hop_list:
+                    next_hop_dict = {}
+                    index_dict = {}
+                    inf_ref = next_hop.get('interface-ref', {})
+                    inf_ref_cfg = inf_ref.get('config', {})
```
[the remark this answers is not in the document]
kerry-meyer: Okay. I missed the fact that the default value setting will prevent an exception if there is no int_ref.

plugins/modules/sonic_static_routes.py
```diff
@@ -0,0 +1,241 @@
...
+                required: True
+                type: dict
+                description:
+                  - An identifier utilised to uniquely reference the next-hop.
+                suboptions:
+                  blackhole:
+                    type: bool
+                    default: False
+                    description:
+                      - Indicates if route is blackholed.
```
kerry-meyer, line 79: This description doesn't add much information.

It would be more useful to define the blackhole buzzword.

For example:

```
Indicates that packets matching this route should be discarded.

File: plugins/modules/sonic_static_routes.py [hunk 9]
```diff
@@ -0,0 +1,241 @@
...
+  state:
+    description:
+      - The state of the configuration after module completion.
+    type: str
+    choices:
+    - merged
+    - deleted
+    default: merged
+"""
+EXAMPLES = """
```
kerry-meyer, line 113: It would be helpful to add a "blackhole" next-hop example, and a merged/modify case that changes an attribute.

The "delete" case would be more realistic if it deleted a single next-hop for the target route (instead of all of them, which is sort of an invalid operation).

tests/regression/roles/sonic_static_routes/defaults/main.yml
```diff
@@ -0,0 +1,191 @@
...
+              - index:
+                  interface: 'Ethernet8'
+                  next_hop: '2.0.0.0'
+          - prefix: '7.0.0.0/8'
+            next_hops:
+              - index:
+                  network_instance: '{{vrf_1}}'
+                  next_hop: '3.0.0.0'
+  - name: test_case_02
+    description: Modify static routes configuration
```
kerry-meyer, line 66: It would be worthwhile to add a case that inserts a new next hop into an existing route.

The corresponding "delete" case, to verify deletion of a next hop by deleting the index for a next hop that still contains other attributes would also be worthwhile. This could be done by leaving at least one tag/track/metric attribute undeleted in test_case_3.

kerry-meyer:
I have reviewed all of the currently posted files and posted my comments.

The only other check remaining is to re-verify test output after any pending second-pass changes have been added.

plugins/module_utils/network/sonic/facts/static_routes/static_routes.py
```diff
@@ -0,0 +1,173 @@
...
+        for static_route in data:
+            static_vrf_dict = {}
+            static_route_list = static_route.get('static', [])
+            vrf_name = static_route.get('vrf', None)
+            static_list = []
+            for static in static_route_list:
+                static_dict = {}
+                prefix = static.get('prefix', None)
+                next_hops = static.get('next-hops', None)
+                next_hop_dict_list = next_hops.get('next-hop', [])
```
kerry-meyer, line 127: Thanks for making the name change to distinguish the two lists of next hops (input list and parsed output dict list), but the usage here is confusing. (It looks like it is reversed from the intended usage.) Please reverse the usage for these two list names ("next_hop_dict_list" and "next_hops_list") to match the naming for the contained items. 

The payload list to which a parsed "next_hop_dict" should be appended is the "next_hop_dict_list".  For the list of next hops obtained from the input playbook, the previously used name "next_hops_list" was fine and should be restored.

tests/regression/roles/sonic_static_routes/defaults/main.yml
```diff
@@ -0,0 +1,181 @@
...
+    input:
+      - vrf_name: 'default'
+        static_list:
+         - prefix: '2.0.0.0/8'
+           next_hops:
+             - index:
+                 interface: 'Ethernet4'
+             - index:
+                next_hop: '3.0.0.0'
+      - vrf_name: '{{vrf_1}}'
```
kerry-meyer, line 154: Deletion of the only remaining next hop for the only remaining prefix in vrf_1 leaves vrf_1 with no prefixes (no routes). As a result, the only deletion at the VRF level (in test_case_05, for vrf 1) is a no-op and doesn't actually test the case of deletion at the vrf level. Instead, since this test is already doing a good job of verifying handling for next hop deletion, it would be better to take this block (next hop deletion for vrf_1) out of test_case_04. This would allow test_case_05 to verify the functionality of vrf level deletion.

kerry-meyer (approved):
The change set, test cases, and test results all look good.

Approved.

PR #97, opened 2022-06-24
Double commit for standard interface naming support bug fix

stalabi1:
##### SUMMARY

I added support for configuration of interfaces in standard naming mode.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

sonic_l2_interfaces


kerry-meyer (approved):
Thanks for doing the double commit.

Approved.

PR #99, opened 2022-07-18
NTP ansible module

mingjunzhang2019:
SUMMARY

Add NTP module

ISSUE TYPE

- New Module Pull Reques

COMPONENT NAME

sonic_ntp

ADDITIONAL INFORMATION

- Regression test log:
[ntp_regression_test.log](https://github.com/ansible-collections/dellemc.enterprise_sonic/files/9463920/ntp_regression_test.log)

- Regression test result: 
[regression-2022-08-31-11-01-14.html.pdf](https://github.com/ansible-collections/dellemc.enterprise_sonic/files/9463926/regression-2022-08-31-11-01-14.html.pdf)

- Negative testing log for server/ntp_keys parameter deleted:
[delete_server_key_parameter_negative_test.log](https://github.com/ansible-collections/dellemc.enterprise_sonic/files/9463930/delete_server_key_parameter_negative_test.log)

- CLI verification show log: 
[ntp_cli_show.txt](https://github.com/ansible-collections/dellemc.enterprise_sonic/files/9135529/ntp_cli_show.txt)

- SONIC DB verification show log:
[ntp_db_show.txt](https://github.com/ansible-collections/dellemc.enterprise_sonic/files/9261400/ntp_db_show.txt)


plugins/modules/sonic_ntp.py
```diff
@@ -0,0 +1,327 @@
...
+#sonic# show ntp global
+#----------------------------------------------
+#NTP Global Configuration
+#----------------------------------------------
+#NTP source-interfaces:  Ethernet0, Ethernet4, Ethernet8, Ethernet16
+#
+#
+# Using merged
+#
+#
```
kerry-meyer, line 269: Please add displaying of before/after state for key configuration for both this "merged" exampel and also for the "deleted" example (above). 

The NTP- specific "show" commands used above for the server and source interface before/after states are good. But, if there is no NTP-specific "show" command that will display the key configuration, please just use "show running-configuration" for the before/after state displays.

plugins/module_utils/network/sonic/argspec/ntp/ntp.py
```diff
@@ -0,0 +1,89 @@
...
+                                   'type': 'int',
+                                   'no_log': True},
+                        'key_type': {'type': 'str',
+                                     'choices': ['NTP_AUTH_SHA1',
+                                                 'NTP_AUTH_MD5',
+                                                 'NTP_AUTH_SHA2_256']},
+                        'key_value': {'type': 'str', 'no_log': True}
+                    },
+                    'type': 'list',
+                    'no_log': True
```
kerry-meyer, line 58: Please remove the "no_log" attribute at the top level for "ntp_keys". It prevents logging of the values that are not "secret". The "key_value" attribute is already protected in this argspec with the "no_log" attribute, so that should be sufficient for providing the needed protection.

plugins/module_utils/network/sonic/config/ntp/ntp.py
```diff
@@ -0,0 +1,550 @@
...
+    normalize_interface_name_list
+)
+from ansible.module_utils.connection import ConnectionError
+
+PATCH = 'PATCH'
+DELETE = 'DELETE'
+
+TEST_KEYS_ADD = [
+    {
+        "vrf": "", "enable_ntp_auth": "",
```
kerry-meyer, line 43: The purpose of the "TEST_KEYS" is to provide the keys that uniquely identify each "list" item that is part of the resource module argument hierarchy (not all of the keys for each list item; only the keys that identify the item).

There is usually only one set of TEST_KEYS (not separate ones for ADD and DEL). 

The format of each dictionary in the list of dictionaries should be:

{"<list_name>":  {"<first_key_name>": "", "<second_key_name>": "", ... }},

In this case, it looks like the required key definitions are present, but there are also extra entries added here that will increase search time during diff calculation.

plugins/module_utils/network/sonic/config/ntp/ntp.py
```diff
@@ -0,0 +1,550 @@
...
+                        server.pop('minpoll')
+                    if not server['maxpoll']:
+                        server.pop('maxpoll')
+
+    def get_create_requests(self, configs, have):
+
+        requests = []
+
+        enable_auth_config = configs.get('enable_ntp_auth', None)
+        if enable_auth_config is not None:
```
kerry-meyer, line 259: Just curious: Is there a reason for this first config type to use:

```
if enable_auth_config is not None:
```
if enable_auth_config is not None:
```

instead of:

```
if enable_auth_config:
```

(The second form is used for the following additional configuration categories.)

File: plugins/module_utils/network/sonic/config/ntp/ntp.py [hunk 5]
```diff
@@ -0,0 +1,550 @@
...
+        existing_ntp_facts = self.get_ntp_facts()
+
+        commands, requests = self.set_config(existing_ntp_facts)
+
+        if commands and len(requests) > 0:
+            if not self._module.check_mode:
+                try:
+                    edit_config(self._module, to_request(self._module, requests))
+                except ConnectionError as exc:
+                    if not re.search("code.*404", str(exc)):
```
kerry-meyer, line 108: Why is "resource not found" excluded from providing a failure message?

plugins/module_utils/network/sonic/config/ntp/ntp.py
```diff
@@ -0,0 +1,550 @@
...
+            commands, requests = self._state_merged(want, have)
+
+        return commands, requests
+
+    def _state_merged(self, want, have):
+        """ The command generator when state is merged
+
+        :param want: the additive configuration as a dictionary
+        :param obj_in_have: the current configuration as a dictionary
+        :rtype: A list
```
kerry-meyer, line 166: This function has no "rtype" parameter, so this line should be deleted.

plugins/module_utils/network/sonic/config/ntp/ntp.py
mingjunzhang2019: This is to check if enable_auth_config is present and enable_auth_config itself may have value True or False.

plugins/module_utils/network/sonic/config/ntp/ntp.py
mingjunzhang2019: This is for DELETE case. It may not be necessary.

plugins/module_utils/network/sonic/config/ntp/ntp.py
```diff
@@ -0,0 +1,550 @@
...
+    def preprocess_delete_commands(self, commands, have):
+
+        new_commands = {}
+
+        enable_auth_command = commands.get('enable_ntp_auth', None)
+        enable_auth_have = have.get('enable_ntp_auth', None)
+        if enable_auth_command is not None and enable_auth_have is not None and \
+           (enable_auth_command or enable_auth_have):
+            new_commands['enable_ntp_auth'] = True
+
```
kerry-meyer, line 305: All of the following attribute pre-processing blocks should only place the command into the new_commands list if the corresponding configuration currently exists on the switch (based on the contents of "have".)

(Alternatively, that check could be done in the "get_xxx_delete_requests" functions, but in that case this function is not needed.)

plugins/module_utils/network/sonic/config/ntp/ntp.py
```diff
@@ -0,0 +1,550 @@
...
+        if trusted_key_command:
+            new_commands['trusted_keys'] = commands['trusted_keys']
+
+        vrf_command = commands.get('vrf', None)
+        if vrf_command:
+            new_commands['vrf'] = commands['vrf']
+
+        return new_commands
+
+    def get_delete_requests(self, configs, have):
```
kerry-meyer, line 328: None of the functions called from this function currently make use of the "have" parameter. If this is not changed (to eliminate the "preprocess_delete_commands" function), the "have" parameter can be deleted for this function and also from the functions that this one calls.

plugins/module_utils/network/sonic/utils/utils.py
```diff
@@ -379,6 +379,16 @@ def normalize_interface_name(configs, module, namekey=None):
                 conf[namekey] = get_normalize_interface_name(conf[namekey], module)
 
 
+def normalize_interface_name_list(configs, module):
```
kerry-meyer, line 382: What is the reason for adding this function? (It apparently isn't being called by the new NTP code.)

tests/regression/roles/sonic_ntp/defaults/main.yml
```diff
@@ -0,0 +1,209 @@
...
+  - name: test_case_19
+    description: Delete NTP authentication keys
+    state: deleted
+    input:
+      ntp_keys:
+        - key_id: 2
+        - key_id: 6
+
+  - name: test_case_20
+    description: Delete NTP authentication - t
```
kerry-meyer, line 201: Is the "- t" intentional? If so, what does it mean?

tests/regression/roles/sonic_ntp/tasks/cleanup_tests.yaml
```diff
@@ -0,0 +1,25 @@
...
+
+- name: Delete MGMT VRF
+  vars:
+    ansible_connection: network_cli
+  sonic_config:
+    commands: "{{ preparations_tests.delete_mgmt_vrf }}"
+  register: output
+  ignore_errors: yes
+
+- name: Change the interface mode to old
```
kerry-meyer, line 17: It would be useful to do something like this in the preparation tests, but not as a cleanup task. (Currently, it is a prerequisite manual task to configure the appropriate naming mode on the switch to match the interface mode specified for the test. Adding an automated execution block like this as a top level test "task" before running any of the requested tests would be the best way to do it.)

To do a cleanup "restore" would require keeping and making use of a persistent state to indicate that the test changed the interface naming state on the switch.

Also, it looks like this "when" clause isn't in the right place for the intended purpose. If it were intended to make execution of the command conditional, it looks like (from the Ansible documentation) it would need to be at the same indentation level as "sonic_config:", just after the "commands: ..." item.

In this case, however, I would recommend deleting this "change the interface mode ..." block.

plugins/module_utils/network/sonic/config/ntp/ntp.py
kerry-meyer: It shouldn't be necessary. If the configuration to be deleted isn't present, no "request" should be generated for that item. If a 404 error occurs, it indicates a code bug that needs to be investigated.

plugins/module_utils/network/sonic/config/ntp/ntp.py
kerry-meyer: Thanks for explaining. After you explained this to me yesterday, I analyzed the case and understood why it is coded this way. I deleted my comment on this line because I realized it was a non-issue.

kerry-meyer:
I have finished checking all changed files and have posted all of my comments for the change set.

plugins/module_utils/network/sonic/config/ntp/ntp.py
```diff
@@ -0,0 +1,550 @@
...
+        elif state == 'merged':
+            commands, requests = self._state_merged(want, have)
+
+        return commands, requests
+
+    def _state_merged(self, want, have):
+        """ The command generator when state is merged
+
+        :param want: the additive configuration as a dictionary
+        :param obj_in_have: the current configuration as a dictionary
```
kerry-meyer, line 165: Please change "obj_in_have" to "have".

plugins/module_utils/network/sonic/config/ntp/ntp.py
```diff
@@ -0,0 +1,550 @@
...
+        return requests
+
+    def preprocess_delete_commands(self, commands, have):
+
+        new_commands = {}
+
+        enable_auth_command = commands.get('enable_ntp_auth', None)
+        enable_auth_have = have.get('enable_ntp_auth', None)
+        if enable_auth_command is not None and enable_auth_have is not None and \
+           (enable_auth_command or enable_auth_have):
```
kerry-meyer, line 303: The deletion should only occur if the two values are the same and is a no-op unless the "and" of these values is True.

Because of this, the line above should be:

```
(enable_auth_command and enable_auth_have)
```

plugins/module_utils/network/sonic/utils/utils.py
mingjunzhang2019: It is used to normalize source-interface list and called in config/ntp.py.

tests/regression/roles/sonic_ntp/defaults/main.yml
mingjunzhang2019: Delete "-t" typo.

plugins/modules/sonic_ntp.py
mingjunzhang2019: Added "before" and "after" states.

plugins/module_utils/network/sonic/argspec/ntp/ntp.py
mingjunzhang2019: If remove "no_log" for "ntp_keys", , ther is sanity error:
ERROR: plugins/modules/sonic_ntp.py:0:0: no-log-needed: Argument 'ntp_keys' in argument_spec could be a secret, though doesn't have `no_log` set found in config

plugins/module_utils/network/sonic/config/ntp/ntp.py
mingjunzhang2019: Removed 404 error exclusion.

plugins/module_utils/network/sonic/facts/ntp/ntp.py
```diff
@@ -0,0 +1,180 @@
...
+
+    def get_ntp_configuration(self):
+        """Get all NTP configuration"""
+
+        config_request = [{"path": "data/openconfig-system:system/ntp/config", "method": GET}]
+        config_response = []
+        try:
+            config_response = edit_config(self._module, to_request(self._module, config_request))
+        except ConnectionError as exc:
+            if re.search("code.*404", str(exc)):
```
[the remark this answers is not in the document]
mingjunzhang2019: "GET" return code 404 (resource not found) when no NTP configuration exists.

plugins/module_utils/network/sonic/config/ntp/ntp.py
```diff
@@ -0,0 +1,550 @@
...
+        :rtype: A list
+        :returns: the commands necessary to merge the provided into
+                  the current configuration
+        """
+        diff = get_diff(want, have, TEST_KEYS_ADD)
+
+        commands = diff
+        requests = []
+        if commands:
+            requests = self.get_create_requests(commands, have)
```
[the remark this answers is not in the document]
mingjunzhang2019: Done

plugins/module_utils/network/sonic/config/ntp/ntp.py
```diff
@@ -0,0 +1,550 @@
...
+            commands = []
+
+        return commands, requests
+
+    def _state_deleted(self, want, have):
+        """ The command generator when state is deleted
+
+        :param want: the objects from which the configuration should be removed
+        :param obj_in_have: the current configuration as a dictionary
+        :param interface_type: interface type
```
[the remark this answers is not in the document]
mingjunzhang2019: Done

plugins/module_utils/network/sonic/config/ntp/ntp.py
```diff
@@ -0,0 +1,550 @@
...
+        :rtype: A list
+        :returns: the commands necessary to remove the current configuration
+                  of the provided objects
+        """
+        diff = get_diff(want, have, TEST_KEYS_DEL)
+
```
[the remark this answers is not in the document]
mingjunzhang2019: Discussed with Kerry: update later.

plugins/module_utils/network/sonic/config/ntp/ntp.py
mingjunzhang2019: Done

plugins/module_utils/network/sonic/config/ntp/ntp.py
mingjunzhang2019: Done

plugins/module_utils/network/sonic/facts/ntp/ntp.py
```diff
@@ -0,0 +1,180 @@
...
+
+        ntp_keys = []
+        if 'openconfig-system:ntp-key' in ntp_keys_response[0][1]:
+            ntp_keys = ntp_keys_response[0][1].get('openconfig-system:ntp-key', {})
+
+        ntp_config = dict()
+        if 'network-instance' in ntp_global_config:
+            ntp_config['vrf'] = ntp_global_config['network-instance']
+        else:
+            ntp_config['vrf'] = None
```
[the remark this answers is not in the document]
mingjunzhang2019: To confirm with input arguments.

plugins/module_utils/network/sonic/utils/utils.py
mingjunzhang2019: Done

tests/regression/roles/sonic_ntp/defaults/main.yml
mingjunzhang2019: Done

plugins/module_utils/network/sonic/config/ntp/ntp.py
mingjunzhang2019: Done

plugins/module_utils/network/sonic/config/ntp/ntp.py
mingjunzhang2019: This is for regression test.

tests/regression/roles/sonic_ntp/tasks/cleanup_tests.yaml
mingjunzhang2019: Removedthe portion.

plugins/module_utils/network/sonic/facts/ntp/ntp.py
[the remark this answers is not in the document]
mingjunzhang2019: Removed the 404 checks.

plugins/module_utils/network/sonic/facts/ntp/ntp.py
```diff
@@ -0,0 +1,180 @@
...
+        :param spec: The facts tree, generated from the argspec
+        :param conf: The configuration
+        :rtype: dictionary
+        :returns: The generated config
+        """
+        return conf
+
+    def get_ntp_configuration(self):
+        """Get all NTP configuration"""
+
```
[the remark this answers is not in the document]
mingjunzhang2019: Done

plugins/module_utils/network/sonic/config/ntp/ntp.py
mingjunzhang2019: It is neded when "have" has "ntp server x.x.x.x minpoll 5 maxpoll 8", then "want" has "ntp server x.x.x.x minpoll 8 maxpoll 10"

plugins/module_utils/network/sonic/config/ntp/ntp.py
mingjunzhang2019: Done

plugins/modules/sonic_ntp.py
kerry-meyer: Thank you.

plugins/module_utils/network/sonic/argspec/ntp/ntp.py
kerry-meyer: Thanks for explaining.

plugins/module_utils/network/sonic/config/ntp/ntp.py
kerry-meyer: After identifying the appropriate list or config element using the TEST_KEYS value or values that uniquely identify a "target" element (in a list or in the config), the "get_diff" function compares the other key/value items to determine if a change is required.

For the example cited, the server IP address alone is sufficient to identify the entry. Once it has been identified, the other key/value pairs (including minpoll and maxpoll are checked to see if they differ between the "have" and "want" dictionaries.

Also, the unique identifying keys for a given item are the same regardless of whether the operation is an ADD or DEL.

plugins/module_utils/network/sonic/config/ntp/ntp.py
kerry-meyer: Verified.

plugins/module_utils/network/sonic/config/ntp/ntp.py
kerry-meyer: Verified.

plugins/module_utils/network/sonic/config/ntp/ntp.py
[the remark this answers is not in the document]
kerry-meyer: Verified.

plugins/module_utils/network/sonic/config/ntp/ntp.py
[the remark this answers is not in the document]
kerry-meyer: Verified.

plugins/module_utils/network/sonic/config/ntp/ntp.py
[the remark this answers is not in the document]
kerry-meyer: It looks like that a bug for which a JIRA should be filed. If you agree, please submit a bug report for it.

In the meantime, it would still be more efficient to handle a "delete_all" by deleting the keys, then doing a top level "delete_all".

plugins/module_utils/network/sonic/config/ntp/ntp.py
kerry-meyer: Pending: We need to revisit the reason for the preprocess_delete_commands function. (See the related comments above.)

plugins/module_utils/network/sonic/facts/ntp/ntp.py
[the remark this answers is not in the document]
kerry-meyer: Verified.

plugins/module_utils/network/sonic/facts/ntp/ntp.py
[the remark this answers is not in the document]
kerry-meyer: Verified.

plugins/module_utils/network/sonic/utils/utils.py
kerry-meyer: I see it now. (I missed it earlier because of the elided diffs for that file.)

tests/regression/roles/sonic_ntp/defaults/main.yml
kerry-meyer: Verified.

tests/regression/roles/sonic_ntp/tasks/cleanup_tests.yaml
kerry-meyer: Verified.

plugins/module_utils/network/sonic/config/ntp/ntp.py
kerry-meyer: Verified.

plugins/module_utils/network/sonic/config/ntp/ntp.py
kerry-meyer: It will work, but it is inefficient because there are more command accesses to the switch than are needed.

plugins/module_utils/network/sonic/facts/ntp/ntp.py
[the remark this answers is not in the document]
kerry-meyer: For each item specified in the input ("want"), a check should be done for the corresponding item in the current switch configuration ("have"). If the item is not there, no "delete" command should be sent to the switch for that item.

plugins/module_utils/network/sonic/config/ntp/ntp.py
kerry-meyer: The regression should pass using the more efficient "and" criterion. (This avoids unnecessary command accesses to the switch.)

plugins/module_utils/network/sonic/config/ntp/ntp.py
mingjunzhang2019: Changed

plugins/module_utils/network/sonic/config/ntp/ntp.py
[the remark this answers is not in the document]
mingjunzhang2019: Added work-around.

plugins/module_utils/network/sonic/config/ntp/ntp.py
kerry-meyer: Verified.

plugins/module_utils/network/sonic/config/ntp/ntp.py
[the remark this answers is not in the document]
kerry-meyer: Verified.

plugins/module_utils/network/sonic/config/ntp/ntp.py
mingjunzhang2019: Removed preprocess_delete_commands function.

plugins/module_utils/network/sonic/config/ntp/ntp.py
mingjunzhang2019: Removed preprocess_delete_commands function.

plugins/module_utils/network/sonic/facts/ntp/ntp.py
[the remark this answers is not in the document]
mingjunzhang2019: Removed those lines.

plugins/module_utils/network/sonic/config/ntp/ntp.py
mingjunzhang2019: Removed preprocess_delete_commands function.

plugins/module_utils/network/sonic/config/ntp/ntp.py
kerry-meyer: Verified.

plugins/module_utils/network/sonic/config/ntp/ntp.py
kerry-meyer: The current parameter usage after the most recent push to this PR looks good.

plugins/module_utils/network/sonic/facts/ntp/ntp.py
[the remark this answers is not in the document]
kerry-meyer: Verified.

plugins/module_utils/network/sonic/config/ntp/ntp.py
kerry-meyer: Verified.

kerry-meyer (approved):

kerry-meyer:
All currently posted changes and test results look good.

Approved.

PR #101, opened 2022-07-22
Added prefix-list attributes to BGP neighbors and peer-groups

stalabi1:
##### SUMMARY

I added the prefix-list related attributes to BGP neighbors and peer-groups

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

sonic_bgp_neighbors
sonic_bgp_neighbors_af

##### ADDITIONAL INFORMATION
[bgp_neighbors_facts_UT.log](https://github.com/ansible-collections/dellemc.enterprise_sonic/files/9169941/bgp_neighbors_facts_UT.log)
[show_running_bgp_pg.txt](https://github.com/ansible-collections/dellemc.enterprise_sonic/files/9169947/show_running_bgp_pg.txt)
[regression-2022-08-03-09-58-29.html.pdf](https://github.com/ansible-collections/dellemc.enterprise_sonic/files/9253768/regression-2022-08-03-09-58-29.html.pdf)
[bgp_neighbors_output.log](https://github.com/ansible-collections/dellemc.enterprise_sonic/files/9253770/bgp_neighbors_output.log)

[bgp_neighbors_af_facts_UT.log](https://github.com/ansible-collections/dellemc.enterprise_sonic/files/9237010/bgp_neighbors_af_facts_UT.log)
[show_running_bgp_nbr.txt](https://github.com/ansible-collections/dellemc.enterprise_sonic/files/9237012/show_running_bgp_nbr.txt)
[regression-2022-08-03-10-17-58.html.pdf](https://github.com/ansible-collections/dellemc.enterprise_sonic/files/9253795/regression-2022-08-03-10-17-58.html.pdf)
[bgp_neighbors_af_output.log](https://github.com/ansible-collections/dellemc.enterprise_sonic/files/9253799/bgp_neighbors_af_output.log)


plugins/module_utils/network/sonic/argspec/bgp_neighbors/bgp_neighbors.py
```diff
@@ -148,6 +148,24 @@ def __init__(self, **kwargs):
                                             },
                                             'type': 'dict'
                                         },
+                                        'ip_afi': {
+                                            'options': {
+                                                'default_policy_name': {'type': 'str'},
```
kerry-meyer, line 153: These default route policy options are also needed for BGP neighbors.

plugins/module_utils/network/sonic/argspec/bgp_neighbors/bgp_neighbors.py
```diff
@@ -148,6 +148,24 @@ def __init__(self, **kwargs):
...
+                                        'prefix_limit': {
+                                            'options': {
+                                                'max_prefixes': {'type': 'int'},
+                                                'prevent_teardown': {'default': False, 'type': 'bool'},
+                                                'warning_threshold': {'type': 'int'},
+                                                'restart_timer': {'type': 'int'}
+                                            },
+                                            'type': 'dict'
+                                        },
+                                        'prefix_list_in': {'type': 'str'},
```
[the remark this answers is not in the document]
stalabi1: bgp neighbors afi safi is in bgp_neighbors_af module so I will be adding all of the afi safi prefix related attributes there

plugins/module_utils/network/sonic/argspec/bgp_neighbors/bgp_neighbors.py
stalabi1: Yes, these attributes will be added to the bgp_neighbors_af module. The PR for that module will be opened today.

plugins/module_utils/network/sonic/argspec/bgp_neighbors/bgp_neighbors.py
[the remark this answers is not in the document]
kerry-meyer: Okay. (We need to fix this resource module structuring in a future release. It will confuse customers.)

For now, I think we can go ahead with getting the 2.0.0 functionality in on with the current structuring.

plugins/module_utils/network/sonic/argspec/bgp_neighbors/bgp_neighbors.py
kerry-meyer: Thanks for explaining. (As mentioned above, the current resource module structuring is confusing, but we can proceed with it for now.)

plugins/module_utils/network/sonic/config/bgp_neighbors/bgp_neighbors.py
```diff
@@ -420,13 +428,44 @@ def build_bgp_peer_groups_payload(self, cmd, have, bgp_as, vrf_name):
...
+                                if afi_safi == 'IPV4_UNICAST':
+                                    if each.get('ip_afi', None) is not None:
+                                        afi_safi_cfg = self.get_afi_safi_config_payload(each['ip_afi'])
+                                        if afi_safi_cfg:
+                                            ip_dict.update({'config': afi_safi_cfg})
+                                    if each.get('prefix_limit', None) is not None:
+                                        pfx_lmt_cfg = self.get_prefix_limit_payload(each['prefix_limit'])
+                                        if pfx_lmt_cfg:
+                                            ip_dict.update({'prefix-limit': {'config': pfx_lmt_cfg}})
+                                    if ip_dict:
```
kerry-meyer, line 451: Since "ip_dict" has been initialized to {}, (which is not "null"), this will always be True. 

I think that the intent of this instruction would be implemented as:

if ip_dict != {}

This same comment applies to the similar cases (above) for afi_safi_cfg and pfx_lmt_cfg, as well as the case below for pfx_lst_cfg.

plugins/module_utils/network/sonic/config/bgp_neighbors/bgp_neighbors.py
```diff
@@ -420,13 +428,44 @@ def build_bgp_peer_groups_payload(self, cmd, have, bgp_as, vrf_name):
...
+                            pfx_lst_cfg = {}
+                            ip_dict = {}
                             if each.get('afi', None) is not None and each.get('safi', None) is not None:
                                 afi_safi = each['afi'].upper() + "_" + each['safi'].upper()
                                 if afi_safi is not None:
                                     afi_safi_name = 'openconfig-bgp-types:' + afi_safi
                                 if afi_safi_name is not None:
                                     samp.update({'afi-safi-name': afi_safi_name})
                                     samp.update({'config': {'afi-safi-name': afi_safi_name}})
+                                if afi_safi == 'IPV4_UNICAST':
```
kerry-meyer, line 442: The following blocks of code could be consolidated to eliminate repeated instructions.

e.g.  (pseudo code), but feel free to depart from this in any way that reduces the repetition:

----------

<store pfx_lmt_cfg>
if afi_safi == 'L2VPN_EVPN':
  <conditional samp.update>
else:
  <common ip handling>
  if ip_dict and if afi_safi == 'IPV4_UNICAST':
    samp.update({'ipv4-unicast': ip_dict})
  if ip_dict and if afi_safi == 'IPV6_UNICAST':
    samp.update({'ipv6-unicast': ip_dict})

tests/regression/roles/sonic_bgp_neighbors/defaults/main.yml
```diff
@@ -80,6 +83,47 @@ deleted_tests:
             v6only: false
 
   - name: test_case_del_02
+    description: Delete BGP peer-group prefix-list attributes
```
kerry-meyer, line 86: Instead of only having a test case that deletes all bottom level attributes for both IP AFI/SAFIs, it would be good to have a test case that verifies deletion of groups of attributes. (ip_afi and prefix_limit). Since the instructions for handling of each of the AFI/SAFIs are the same, it would be adequate to only test bottom level attribute deletion for one AFI/SAFI and to do group deletion for the other one.

plugins/module_utils/network/sonic/config/bgp_neighbors/bgp_neighbors.py
```diff
@@ -797,6 +881,59 @@ def delete_specific_peergroup_param_request(self, vrf_name, cmd):
...
+        requests = []
+        default_policy_name = ip_afi.get('default_policy_name', None)
+        send_default_route = ip_afi.get('send_default_route', None)
+        if default_policy_name:
+            delete_path = delete_static_path + '/afi-safis/afi-safi=%s/%s/config/default-policy-name' % (afi_safi_name, afi_safi)
+            requests.append({'path': delete_path, 'method': DELETE})
+        if send_default_route:
+            delete_path = delete_static_path + '/afi-safis/afi-safi=%s/%s/config/send_default_route' % (afi_safi_name, afi_safi)
+            requests.append({'path': delete_path, 'method': DELETE})
+
```
kerry-meyer, line 916: Don't we need handling for bulk deletion of the whole ip_afi "group"? (This same question applies to "delete_prefix_limit_requests".)

plugins/module_utils/network/sonic/config/bgp_neighbors/bgp_neighbors.py
```diff
@@ -484,6 +533,36 @@ def build_bgp_peer_groups_payload(self, cmd, have, bgp_as, vrf_name):
         payload = {'openconfig-network-instance:peer-groups': {'peer-group': bgp_peer_group_list}}
         return payload, requests
 
+    def get_afi_safi_config_payload(self, ip_afi):
```
kerry-meyer, line 536: Please rename this function to reduce confusion with the higher (generic) level afi_safi category. It would be good to explicitly include "ip" in the name to clarify that fact that it is to be used only for the IP specific afi_safi cases.

e.g. get_ip_afi_config_payload or get_ip_afi_safi_config_payload

plugins/module_utils/network/sonic/config/bgp_neighbors/bgp_neighbors.py
kerry-meyer: Further investigation into this question has shown that there is no good way, with the current Ansible infrastructure, to use a YAML playbook to request deletion of a dictionary that is not a list item. This is because there is no way to differentiate, in the command list returned by the "AnsibleModule" method, between a dictionary that is not specified at all in the input playbook versus a dictionary for which only the name is specified in the playbook: Both cases yield "'dict_name': None" with the yaml.safe_load() method and a "null" value in the parsed input command list.

In addition, currently existing resource module implementations are not providing the ability to delete a dictionary that is not a list item.

For these reasons, I am dropping this request.

tests/regression/roles/sonic_bgp_neighbors/defaults/main.yml
kerry-meyer: See comments on the related issue of "bulk deletion" of the contents of a dictionary that is not a list item: Because this is not a conveniently supported capability and no support for it is currently provided in other resource modules, I am dropping this test case request.

plugins/module_utils/network/sonic/config/bgp_neighbors_af/bgp_neighbors_af.py
```diff
@@ -73,6 +73,14 @@ class Bgp_neighbors_af(ConfigBase):
     allowas_origin_path = "/allow-own-as/config/origin"
     allowas_value_path = "/allow-own-as/config/as-count"
     allowas_enabled_path = "/allow-own-as/config/enabled"
+    prefix_list_in_path = "/prefix-list/config/import-policy"
+    prefix_list_out_path = "/prefix-list/config/export-policy"
+    df_policy_name_path = "/%s/config/default-policy-name"
```
kerry-meyer, line 78: Please use "def" instead of "df" for this abbreviation. ("DF" has a protocol meaning: "Designated Forwarder" and it might confuse readers when they see it used as an abbreviation for "default".)

shatalabi: Done

plugins/module_utils/network/sonic/config/bgp_neighbors_af/bgp_neighbors_af.py
```diff
@@ -220,6 +228,36 @@ def get_allowas_in(self, match, conf_neighbor_val, conf_afi, conf_safi):
                             mat_allowas_in = mat_nei_addr_fam.get('allowas_in', None)
         return mat_allowas_in
 
+    def get_ip_afi_cfg_payload(self, ip_afi):
```
kerry-meyer, line 231: Since the logic of get_ip_afi_cfg_payload and get_prefix_limit_payload is identical in the bgp_neighbors and bgp_neighbors_af "config" files, it would be better to locate a single set of these two functions in the ".../bgp_utils.py" file and invoke the functions from each of the resource modules that use them. Use of the "ip_afi_cfg" local variable here is a better name than the one used for "bgp_neighbors", so the versions shown here would be good candidates for copying (moving) into the bgp_utils.py file.

plugins/module_utils/network/sonic/config/bgp_neighbors_af/bgp_neighbors_af.py
```diff
@@ -425,6 +530,22 @@ def process_delete_specific_params(self, vrf_name, conf_neighbor_val, conf_nei_a
                         if value is not None:
                             if self.append_delete_request(requests, value, mat_alllowas_in, 'value', url, self.allowas_value_path):
                                 self.append_delete_request(requests, True, {'enabled': True}, 'enabled', url, self.allowas_enabled_path)
+
+                mat_ip_afi = mat_nei_addr_fam.get('ip_afi', None)
+                mat_prefix_limit = mat_nei_addr_fam.get('prefix_limit', None)
```
kerry-meyer, line 535: It looks like this block could be restructured as follows to reduce repetition of instructions and to eliminate unnecessary conditional branching:

```
                if conf_prefix_limit and mat_prefix_limit:
                    requests.extend(self.delete_prefix_limit_requests(conf_prefix_limit, mat_prefix_limit, conf_afi_safi_val, url)
                if conf_ip_afi and mat_ip_afi:
                    requests.extend(self.delete_ip_afi_requests(conf_ip_afi, mat_ip_afi, conf_afi_safi_val, url))
```
Note: This comment takes into account the fact that the append_delete_request function, invoked for these cases, will proceed only if the specified configuration (based on the key) is already present on the device for the specified afi/safi neighbor.

plugins/module_utils/network/sonic/config/bgp_neighbors_af/bgp_neighbors_af.py
```diff
@@ -425,6 +530,22 @@ def process_delete_specific_params(self, vrf_name, conf_neighbor_val, conf_nei_a
...
+                if conf_afi_safi_val == 'ipv4-unicast':
```
kerry-meyer, line 536: It looks like this block could be restructured as follows to reduce repetition of instructions and to eliminate unnecessary conditional branching:

```
                if conf_prefix_limit and mat_prefix_limit:
                    requests.extend(self.delete_prefix_limit_requests(conf_prefix_limit, mat_prefix_limit, conf_afi_safi_val, url)
                if conf_ip_afi and mat_ip_afi:
                    requests.extend(self.delete_ip_afi_requests(conf_ip_afi, mat_ip_afi, conf_afi_safi_val, url))
```
Note: This comment takes into account the fact that the append_delete_request function, invoked for these cases, will proceed only if the specified configuration (based on the key) is already present on the device for the specified afi/safi neighbor.

stalabi1: Right, obvious.

kerry-meyer (approved):

plugins/module_utils/network/sonic/config/bgp_neighbors_af/bgp_neighbors_af.py
kerry-meyer: The requested change has been committed. This issue is resolved.

plugins/module_utils/network/sonic/config/bgp_neighbors_af/bgp_neighbors_af.py
kerry-meyer: The requested changes have been committed. This issue is resolved.

kerry-meyer (approved):
The current version of the change set looks good for merging.

Approved.

PR #106, opened 2022-09-01
Change NTP get fact on NTP servers

mingjunzhang2019:
SUMMARY

get-fact on NTP server brings in unsupported paramer. This is to fix the issue.

ISSUE TYPE

- Bugfix Pull Request

COMPONENT NAME

sonic_ntp

ADDITIONAL INFORMATION

regression test log: 
[ntp_regression_log.log](https://github.com/ansible-collections/dellemc.enterprise_sonic/files/9472369/ntp_regression_log.log)

regression test result:
[regression-2022-09-01-10-10-23.html.pdf](https://github.com/ansible-collections/dellemc.enterprise_sonic/files/9472376/regression-2022-09-01-10-10-23.html.pdf)

### DIFF
--- a/plugins/module_utils/network/sonic/facts/ntp/ntp.py
+++ b/plugins/module_utils/network/sonic/facts/ntp/ntp.py
@@ -128,10 +128,12 @@ class NtpFacts(object):
         servers = []
         for ntp_server in ntp_servers:
             if 'config' in ntp_server:
-                server = ntp_server['config']
-                if 'key-id' in server:
-                    server['key_id'] = server['key-id']
-                    server.pop('key-id')
+                server = {}
+                server['address'] = ntp_server['config'].get('address', None)
+                if 'key-id' in ntp_server['config']:
+                    server['key_id'] = ntp_server['config']['key-id']
+                server['minpoll'] = ntp_server['config'].get('minpoll', None)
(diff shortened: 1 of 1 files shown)


plugins/module_utils/network/sonic/facts/ntp/ntp.py
```diff
@@ -128,10 +128,12 @@ def get_ntp_configuration(self):
...
         for ntp_server in ntp_servers:
             if 'config' in ntp_server:
-                server = ntp_server['config']
-                if 'key-id' in server:
-                    server['key_id'] = server['key-id']
-                    server.pop('key-id')
+                server = {}
+                server['address'] = ntp_server['config'].get('address', None)
+                if 'key-id' in ntp_server['config']:
+                    server['key_id'] = ntp_server['config']['key-id']
```
kerry-meyer, line 134: Please modify the logic here to only set the minpoll and maxpoll values into the config if the user has requested that values be set.

e.g.

```
if 'minpoll' in ntp_server['config']:
    minpoll = ntp_server['config'].get('minpoll', None)
    if minpoll is not None:
    server['minpoll'] = minpoll
```
This will avoid the possibility of putting "None" values into the config request.

kerry-meyer (approved):
The code changes and test results look good.

Thanks for the quick action on this fix.

Approved.

PR #107, opened 2022-09-02
Change NTP key values in regression test cases

mingjunzhang2019:
SUMMARY
NTP encrypted key value is switch dependent. So change it to use those generated with regression switch.

ISSUE TYPE

- Regression test case modification Pull Request

ADDITIONAL INFORMATION

- regression test log:
[ntp_regression_test.log](https://github.com/ansible-collections/dellemc.enterprise_sonic/files/9474189/ntp_regression_test.log)

- regression test result: 
[regression-2022-09-01-18-03-17.html.pdf](https://github.com/ansible-collections/dellemc.enterprise_sonic/files/9474192/regression-2022-09-01-18-03-17.html.pdf)

### DIFF
--- a/tests/regression/roles/sonic_ntp/defaults/main.yml
+++ b/tests/regression/roles/sonic_ntp/defaults/main.yml
@@ -149,11 +149,11 @@ tests:
       ntp_keys:
         - key_id: 2
           key_type: NTP_AUTH_SHA1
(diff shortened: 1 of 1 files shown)


kerry-meyer (approved):
Thanks for making these changes.

Approved.

PR #110, opened 2022-09-06
Fixed regex to support standard naming with port breakout mode and resolved other minor bugs

stalabi1:
- Fixed the regex expression in sonic.py to support standard naming with port breakout mode
- Removed AutoEscapeExtension in config since it is now built-in with Jinja 3.10

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME
sonic.py
bgp_communities
bgp_ext_communities
lag_interfaces
interfaces_util

##### ADDITIONAL INFORMATION
Executed in standard naming mode :
[regression-2022-09-06-12-58-22.html.pdf](https://github.com/ansible-collections/dellemc.enterprise_sonic/files/9500415/regression-2022-09-06-12-58-22.html.pdf)

test regression results for l2_interfaces after reverting standard naming bug fix :
[regression-2022-09-07-11-39-13.html.pdf](https://github.com/ansible-collections/dellemc.enterprise_sonic/files/9508797/regression-2022-09-07-11-39-13.html.pdf)

### DIFF
--- a/.github/workflows/ansible-test.yml
+++ b/.github/workflows/ansible-test.yml
@@ -27,10 +27,10 @@ jobs:
         with:
           path: ansible_collections/dellemc/enterprise_sonic
 
-      - name: Set up Python 3.8
+      - name: Set up Python 3.9
         uses: actions/setup-python@v1
         with:
-          python-version: 3.8
+          python-version: 3.9
 
       - name: Install ansible-base (${{ matrix.ansible }})
         run: pip install https://github.com/ansible/ansible/archive/${{ matrix.ansible }}.tar.gz --disable-pip-version-check
--- a/plugins/action/sonic.py
+++ b/plugins/action/sonic.py
@@ -36,8 +36,7 @@ version_added: 1.0.0
 
 class ActionModule(ActionNetworkModule):
 
-    def run(self, tmp=None, task_vars=None):
-        del tmp  # tmp no longer has any effect
+    def run(self, task_vars=None):
 
         module_name = self._task.action.split('.')[-1]
         self._config_module = True if module_name == 'sonic_config' else False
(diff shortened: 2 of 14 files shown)


kerry-meyer (commented):
The currently posted changes and test results look good, but I have a few questions about the port:

1) Are the posted test results for a run with standard interface-naming configured? If so, please state that in the description. If not, it is important to verify this to confirm that the fix works, at least for the port breakout test.
2) The following files had changes posted in the corresponding "main" branch PR that aren't present in this change set. Is there a reason why these changes were not included here?

  - plugins/action/sonic.py  (Delete the section related to the obsolete "tmp" object.)
   - plugins/module_utils/network/sonic/config/l2_interfaces/l2_interfaces.py (Back out the standard-interface naming fix that is no longer needed because of the new fix.)
  - plugins/module_utils/network/sonic/config/system/system.py (default entry handling)
  
3) Is there a reason why the regression tests for the "system" resource module are different? (Why does the one on the "main" branch delete the ipv4 bool for anycast address while the 1.x version of that test does not?)

stalabi1:
@kerry-meyer  

Yes, the regression was executed in standard naming mode. I didn't change sonic.py because the sanity check didn't throw error like it did for the 2.x branch. I didn't make any system changes because the sonic code for system is different between 3.x and 4.x which is why there are no system failures for 1.x regression. Do you still want me to update those things anyways?

kerry-meyer:
Thanks for explaining.

The additional pushes done today should resolve my only remaining concerns. (No additional changes are needed.)

kerry-meyer (approved):
The currently posted changes and test results look good.

Thanks for porting these fixes.

Approved.

PR #118, opened 2022-09-14
Add server "prefer" attribute to NTP module

mingjunzhang2019:
SUMMARY
This is to add NTP server "prefer" attribute to NTP module.

ISSUE TYPE
- Module Enhancement Pull Request

COMPONENT NAME
sonic_ntp

ADDITIONAL INFORMATION
- regression test log: 
[regression_test.log](https://github.com/ansible-collections/dellemc.enterprise_sonic/files/9767805/regression_test.log)

- regression test result: 
[regression-2022-10-12-10-50-29.html.pdf](https://github.com/ansible-collections/dellemc.enterprise_sonic/files/9767807/regression-2022-10-12-10-50-29.html.pdf)

### DIFF
--- a/tests/regression/roles/sonic_ntp/defaults/main.yml
+++ b/tests/regression/roles/sonic_ntp/defaults/main.yml
@@ -81,6 +81,7 @@ tests:
         - address: "{{ ntp_ip_server_3 }}"
           minpoll: 7
           maxpoll: 10
+          prefer: true
 
   - name: test_case_06
     description: Delete a NTP source interface
--- a/tests/regression/roles/sonic_ntp/defaults/main.yml
+++ b/tests/regression/roles/sonic_ntp/defaults/main.yml
@@ -84,20 +84,30 @@ tests:
           prefer: true
 
   - name: test_case_06
+    description: Configure NTP server prefer to false
+    state: merged
+    input:
+      servers:
+        - address: "{{ ntp_ip_server_3 }}"
+          minpoll: 6
+          maxpoll: 10
+          prefer: false
+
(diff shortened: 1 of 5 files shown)


tests/regression/roles/sonic_ntp/defaults/main.yml
```diff
@@ -81,6 +81,7 @@ tests:
         - address: "{{ ntp_ip_server_3 }}"
           minpoll: 7
           maxpoll: 10
+          prefer: true
 
   - name: test_case_06
```
kerry-meyer, line 86: Please add a test case to change the value of the "prefer" bool. (Changing it from "true" to "false" is sufficient. This is to verify that the update logic in the config file correctly handles the case of changing this value on an already existing server configuration.

kerry-meyer (changes requested):
This change set looks good: All "functional" code changes look fine.

I have posted only one comment, inline, in the "tests/regression/roles/sonic_ntp/defaults/main.yml" file. (This is regarding an additional test case to verify that changing the value of the "prefer" bool works correctly).

tests/regression/roles/sonic_ntp/defaults/main.yml
mingjunzhang2019: Add more regression test cases. Thanks

kerry-meyer: The additional coverage and corresponding test result looks good.

Thanks for adding this case.

kerry-meyer (approved):
All changes and test results look good.

Approved.

PR #138, opened 2022-11-17
Add Enterprise SONiC "vrrp" resource module

santhosh-kt:
##### SUMMARY

- Add Enterprise SONiC 'VRRP' resource module:
    - Provide data definitions in 'argspec'.
    - Implement 'facts' to gather facts for the module.
    - Implement 'config' to apply configurations for `merged` , `deleted`, `overridden` and `replaced` states.
- Define regression test cases to verify correct functionality of vrrp resource module.
##### ISSUE TYPE

- New Module Pull Request

##### COMPONENT NAME

- sonic_vrrp

#### Test Results
- Regression test HTML report:
[vrrp_regression_report.pdf](https://github.com/ansible-collections/dellemc.enterprise_sonic/files/10034952/vrrp_regression_report.pdf)

- Regression test detailed log:
[vrrp_regression_terminal_output.txt](https://github.com/ansible-collections/dellemc.enterprise_sonic/files/10034956/vrrp_regression_terminal_output.txt)

- Show running config and respective facts gathered:
[show_running_get_facts_output.txt](https://github.com/ansible-collections/dellemc.enterprise_sonic/files/10034957/show_running_get_facts_output.txt)


plugins/module_utils/network/sonic/config/vrrp/vrrp.py
```diff
@@ -0,0 +1,686 @@
...
+from ansible_collections.dellemc.enterprise_sonic.plugins.module_utils.network.sonic.sonic import (
+    to_request,
+    edit_config
+)
+
+PATCH = 'patch'
+DELETE = 'delete'
+
+ipv4_path = '/openconfig-if-ip:ipv4/addresses/address=1.1.1.1/'
+ipv6_path = '/openconfig-if-ip:ipv6/addresses/address=1::1/'
```
stalabi1, line 40: Why are these addresses hardcoded?

stalabi1:
For the regression test detailed log you posted a link to the PR. Please upload the txt file.

santhosh-kt:
Last upload must have failed. Now uploaded correct regression detailed log file.

plugins/module_utils/network/sonic/config/vrrp/vrrp.py
santhosh-kt: The VRRP REST API needs an IP address as default in the URL. On comparing with sonic-cli, there they use these hardcoded addresses to fetch the request.

santhosh-kt: It acts as a placeholder value in the URL. Except 127.0.0.1, any IP address works on the URL.

stalabi1: I understand your reasoning. I have experienced similar cases where default vrf was hardcoded from the sonic-cli code although other vrf instances were permitted. In those cases, I would add a vrf attribute to the argspec and make it required or set a default value. I recommend that you add an address attribute to the argspec and specify it as required. Then you could do a regex check in config to determine if it is ipv4 or ipv6.

santhosh-kt: Thats an interesting solution however we are using "afi" attribute to determine if its a ipv4 or ipv6 similar to BGP. Based on that attribute, we are hardcoding "1.1.1.1" or "1::1".

stalabi1: I recommended this solution because you stated that "any IP address works". I will find out if being able to specify the address has any useful functional flexibility and get back to you.

santhosh-kt:
Closing this PR.

Raised a new PR with unit testcases - https://github.com/ansible-collections/dellemc.enterprise_sonic/pull/353

PR #182, opened 2023-03-13
BGP Unit test code added

kannankvs:
##### SUMMARY
Adding unit tests code for BGP module

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
BGP Unit Test Module

### DIFF
--- a/tests/unit/modules/network/sonic/fixtures/sonic_bgp.yaml
+++ b/tests/unit/modules/network/sonic/fixtures/sonic_bgp.yaml
@@ -1,0 +1,143 @@
+---
+merged_01:
+  module_args:
+    config:
+      - bgp_as: 4
+        router_id: 10.2.2.4
+        log_neighbor_changes: False
+        timers:
+          holdtime: 20
+          keepalive_interval: 30
+        bestpath:
+          as_path:
+            confed: True
+            ignore: True
+            multipath_relax: False
+            multipath_relax_as_set: True
+          compare_routerid: True
+          med:
+            confed: True
+            missing_as_worst: True
+            always_compare_med: True
+        max_med:
+          on_startup:
+            timer: 667
+            med_val: 7878
+  existing_bgp_config:
+    - path: "/data/openconfig-network-instance:network-instances/network-instance=default/protocols/protocol=BGP,bgp/bgp/global"
+      response:
+        code: 200
+        value:
+          openconfig-network-instance:global:
+            config:
+              as: 4
+    - path: "data/sonic-vrf:sonic-vrf/VRF/VRF_LIST"
+      response:
+        code: 200
+        value:
+          sonic-vrf:VRF_LIST:
+            - vrf_name: default
+  expected_config_requests:
+    - path: "/data/openconfig-network-instance:network-instances/network-instance=default/protocols/protocol=BGP,bgp/bgp/global/config"
+      method: "patch"
+      data:
+        openconfig-network-instance:config:
+          router-id: "10.2.2.4"
+          as: 4.0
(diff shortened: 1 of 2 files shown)


stalabi1 (approved):

tests/unit/modules/network/sonic/fixtures/sonic_bgp.yaml
```diff
@@ -0,0 +1,143 @@
...
+            med-missing-as-worst: True
+            always-compare-med: True
+    - path: "/data/openconfig-network-instance:network-instances/network-instance=default/protocols/protocol=BGP,bgp/bgp/global/use-multiple-paths/ebgp/config"
+      method: "patch"
+      data:
+        openconfig-network-instance:config:
+          as-set: True
+deleted_01:
+  module_args:
+    config:
```
kerry-meyer, line 80: There are several attributes for which deletion is not being tested here. Is there a reason not to test deletion for these attributes?

- timers
- max_med
- bestpath/as_path/multipath_relax_as_set
- med/always_compare_med

tests/unit/modules/network/sonic/fixtures/sonic_bgp.yaml
kerry-meyer, line 1: The "delete_all" case is currently not being tested in this file. If the "merged_01" case were duplicated/repeated as "merged_02", it could be followed by a "delete_all" test ("deleted_02"?) to test that code path.

kerry-meyer (changes requested):
There are a few additional test cases or test case attributes that would help to improve the UT coverage for this resource module. I have posted an inline comment and a "file" comment asking about the possibility of adding these:

  - additional "delete" attributes
  - testing of the "delete all" case.

kannankvs:
@kerry-meyer : The plan for UT is as follows. At first, we will write UT cases to meet 70% coverage. This is to put a TICK MARK against the OneAnsible requirement for UT. Later on, we need to plan to add more test cases to increase the coverage. Hence, I suggest to keep the above comments as FutureEnhancements and merge the PR with the currently available test cases.

kerry-meyer:
These first few UT PRs will be used by others as examples. It would be helpful if they at least did merge/delete for each defined attributed plus a "delete all" case. This should be the minimum requirement for each resource module. (I understand that each module may have some other special case branching that is beyond the scope of the initial UT implementation. But the items mentioned here seem like the basics, common to all resource modules.)

If there is a specific reason for this resource module's initial UT to omit the items mentioned in these comments for now, please explain the reason. (e.g. Why delete most of the attributes, but skip the ones mentioned above? If there are some special circumstances, we can consider deferring these items for now.)

kannankvs:
@kerry-meyer : Added the test cases that you mentioned.

tests/unit/modules/network/sonic/fixtures/sonic_bgp.yaml
kerry-meyer: Thanks for adding these attributes to the test.

Resolved.

tests/unit/modules/network/sonic/fixtures/sonic_bgp.yaml
kerry-meyer: Thanks for adding this test case.

Resolved.

kerry-meyer (approved):
Thank you for adding the additional requested test cases for this PR.

Note: The UT implementation in this resource module provides a good base example. For the remaining group of initial UT PRs, 70% coverage for the "target" resource module will be a sufficient UT criterion for approval and merging.

Approved.

PR #199, opened 2023-03-31
Install dependencies from ansible galaxy in code coverage workflow

ArunSaravananBalachandran:
##### SUMMARY

Install released version of dependencies from ansible galaxy in code coverage workflow, instead of installing directly from git.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

- workflow


kerry-meyer (approved):
Thank you for your quick action on fixing this problem.

Approved.

PR #201, opened 2023-03-31
Fixed bgp af bug

stalabi1:
##### SUMMARY

Added function calls to modify vni and route advertise for a single bgp af

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

sonic_bgp_af

##### ADDITIONAL INFORMATION
[regression-2023-03-31-12-20-24.html.pdf](https://github.com/ansible-collections/dellemc.enterprise_sonic/files/11127307/regression-2023-03-31-12-20-24.html.pdf)

[bgp_af_reg_output.log](https://github.com/ansible-collections/dellemc.enterprise_sonic/files/11123880/bgp_af_reg_output.log)


plugins/module_utils/network/sonic/config/bgp_af/bgp_af.py
kerry-meyer, line 1: These changes look fine for fixing the problem, but please add/modify the regression test file to provide coverage for the case where the problem occurred.

kerry-meyer (approved):
Thanks for the additional changes and the UT fixes.

Approved.

PR #222, opened 2023-04-04
Unit testing code for modules/sonic_facts.py added

kannankvs:
##### SUMMARY
Unit testing code for modules/sonic_facts.py added

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
sonic_facts Unit Testing

##### DESCRIPTION
The main objective of UT is to ensure that all files present in plugins/modules folder are covered. This particular file sonic_facts.py is a common module to pull out the facts for all other modules.
This file is not specific to any particular protocol and hence the unit test cases for this file does not follow the same strategy as the protocol unit testing code. This unit testing is only to ensure that the code coverage for this file is met.
The actual functionality for each protocol is being validated in the unit testing that corresponds to the protocol unit testing code. 
The strategy used in this proposed change is to call any one protocol facts code and test the same using stub return values. 
The return values from stub does not matter since this particular file does not validate the return values.
i.e. The main agenda for this unit testing is only to test whether the code flow for sonic_facts works as expected which is being testing using the UT code.

### DIFF
--- a/tests/unit/modules/network/sonic/fixtures/sonic_facts.yaml
+++ b/tests/unit/modules/network/sonic/fixtures/sonic_facts.yaml
@@ -1,0 +1,5 @@
+---
+merged_01:
+  module_args:
+    gather_network_resources:
+      - "vlans"
--- a/tests/unit/modules/network/sonic/test_sonic_facts.py
+++ b/tests/unit/modules/network/sonic/test_sonic_facts.py
@@ -1,0 +1,45 @@
+from __future__ import absolute_import, division, print_function
+
+__metaclass__ = type
+
+from ansible_collections.dellemc.enterprise_sonic.tests.unit.compat.mock import (
+    patch,
+)
+from ansible_collections.dellemc.enterprise_sonic.plugins.modules import (
+    sonic_facts,
+)
+from ansible_collections.dellemc.enterprise_sonic.tests.unit.modules.utils import (
+    set_module_args,
+)
+from .sonic_module import TestSonicModule
+
+
+class TestSonicInterfacesModule(TestSonicModule):
+    module = sonic_facts
+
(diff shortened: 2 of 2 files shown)


stalabi1 (approved):

kerry-meyer (commented):
In the "Description" section for this PR, please provide some description of the idea behind this UT PR. More specifically:

- Briefly explain the considerations and objectives for testing the functionality (and also for providing code coverage) for the "facts" module and why the test structure differs so much from the other UT code sets.
- Explain how the UT proposed here verifies correct functionality of the facts module. (If it doesn't do that, is there a reason why it is useful? If so, what is the reason.)
- Explain the strategy used for achieving the testing objectives.
- Explain the reason for specifying that the facts gathered are for the "vlans" module when the simulated return data is for BGP. (I understand that it may not matter, but shouldn't they be consistent?)

kannankvs:
> In the "Description" section for this PR, please provide some description of the idea behind this UT PR. More specifically:
> 
> * Briefly explain the considerations and objectives for testing the functionality (and also for providing code coverage) for the "facts" module and why the test structure differs so much from the other UT code sets.
> * Explain how the UT proposed here verifies correct functionality of the facts module. (If it doesn't do that, is there a reason why it is useful? If so, what is the reason?)
> * Explain the strategy used in this proposed change set for achieving the testing objectives.
> * Explain the reason for specifying that the facts gathered are for the "vlans" module when the simulated return data is for BGP. (I understand that it may not matter, but shouldn't they be consistent?)

@kerry-meyer : I have added some description to the PR. Kindly check and let me know if anything else is required. We can get on to a call if there are further comments.

kerry-meyer (approved):
Thank you for providing the explanation of the test strategy.

Approved.

PR #226, opened 2023-04-07
Unit testing for logging module added

kannankvs:
##### SUMMARY
Unit testing for logging module added

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
logging Unit Testing

### DIFF
--- a/tests/unit/modules/network/sonic/sonic_module.py
+++ b/tests/unit/modules/network/sonic/sonic_module.py
@@ -113,22 +113,6 @@ class TestSonicModule(ModuleTestCase):
 
         return responses
 
-    def send_requests_side_effect(self, module, commands):
-        """Side effect function for 'send' requests mock"""
-        responses = []
-        for command in commands:
-            response = []
-            path = update_url(command['path'])
-            method = command['method'].lower()
-            if 'data' in command:
-                data = command['data']
-            else:
-                data = None
-
-            self.config_requests_sent.append({'path': path, 'method': method, 'data': data})
-
-        return responses
-
     def execute_module(self, failed=False, changed=False):
         if failed:
             result = self.failed()
--- a/tests/unit/modules/network/sonic/sonic_module.py
+++ b/tests/unit/modules/network/sonic/sonic_module.py
@@ -113,6 +113,22 @@ class TestSonicModule(ModuleTestCase):
 
         return responses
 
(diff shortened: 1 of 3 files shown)


tests/unit/modules/network/sonic/sonic_module.py
kerry-meyer, line 1: Why does this function become necessary with the addition of the "logging" UT? (What does it do? Why wasn't it needed for UT in other resource modules?

tests/unit/modules/network/sonic/test_sonic_logging.py
```diff
@@ -0,0 +1,87 @@
...
+    def setUp(self):
+        super(TestSonicInterfacesModule, self).setUp()
+        self.facts_edit_config = self.mock_facts_edit_config.start()
+        self.config_edit_config = self.mock_config_edit_config.start()
+        self.facts_edit_config.side_effect = self.facts_side_effect
+        self.config_edit_config.side_effect = self.config_side_effect
+        self.get_interface_naming_mode = self.mock_get_interface_naming_mode.start()
+        self.get_interface_naming_mode.return_value = 'standard'
+        self.utils_send_requests = self.mock_utils_send_requests.start()
+        self.utils_send_requests.side_effect = self.send_requests_side_effect
```
kerry-meyer, line 45: Related to the question asked on the sonic_module file: What is the purpose of this additional setup step?

kerry-meyer (commented):
I have posted a couple of questions about the addition of the "send_requests_side_effect" function in the UT for this resource module. I am not expecting these to result in change requests, but would like to understand the functionality before approving and merging this PR.

kannankvs:
@kerry-meyer : Modified the mocking method for send_requests; reused config_side_effect instead of introducing new mocking method. Previously, I was thinking that mocking send_requests differently might be required in many modules. Its now removed and same old config_side_effect is reused.

tests/unit/modules/network/sonic/sonic_module.py
kerry-meyer: This file has been deleted from the change set.

I am marking this issue "Resolved".

tests/unit/modules/network/sonic/test_sonic_logging.py
kerry-meyer: This code has been revised.

I am marking this issue "Resolved".

PR #234, opened 2023-04-13
Enterprise SONIC Updated Port_breakout with "replaced" and "overridden" options

divya-balasubramania:
##### SUMMARY
"Replaced" and "overridden" states support added for port_breakout

##### COMPONENT NAME
port_breakout

##### OUTPUT
Test execution detailed logs -  [port_breakout_execution.txt](https://github.com/ansible-collections/dellemc.enterprise_sonic/files/11220344/port_breakout_execution.txt)
Test Report - [Port_breakout_regression.pdf](https://github.com/ansible-collections/dellemc.enterprise_sonic/files/11220351/Port_breakout_regression.pdf)

### DIFF
--- a/plugins/module_utils/network/sonic/config/port_breakout/port_breakout.py
+++ b/plugins/module_utils/network/sonic/config/port_breakout/port_breakout.py
@@ -207,13 +207,9 @@ class Port_breakout(ConfigBase):
         commands = []
         requests = []
         if not diff:
-              return commands,requests
+            return commands, requests
         commands = have
         requests = self.get_delete_port_breakout_requests(commands, have)
-
-        import sys
-        #with open('requests', 'w') as sys.stdout:
-        #    print(requests)
 
         if commands and len(requests) > 0:
             send_requests(self._module, requests)
@@ -221,8 +217,6 @@ class Port_breakout(ConfigBase):
             commands = []
 
         commands = diff
-        #with open('diff', 'w') as sys.stdout:
-        #    print(diff)
 
         requests = self.get_modify_port_breakout_requests(commands, want)
         if commands and len(requests) > 0:
--- a/plugins/module_utils/network/sonic/config/port_breakout/port_breakout.py
+++ b/plugins/module_utils/network/sonic/config/port_breakout/port_breakout.py
@@ -27,6 +27,7 @@ from ansible_collections.dellemc.enterprise_sonic.plugins.module_utils.network.s
 from ansible_collections.dellemc.enterprise_sonic.plugins.module_utils.network.sonic.utils.utils import (
     update_states,
     get_diff,
+    send_requests,
     get_speed_from_breakout_mode,
     get_breakout_mode,
 )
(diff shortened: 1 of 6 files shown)


plugins/module_utils/network/sonic/config/port_breakout/port_breakout.py
```diff
@@ -175,6 +176,56 @@ def _state_deleted(self, want, have, diff):
...
+
+        :param want: the additive configuration as a dictionary
+        :param obj_in_have: the current configuration as a dictionary
+        :rtype: A list
+        :returns: the commands necessary to merge the provided into
+                  the current configuration
+        """
+        commands = []
+        requests = []
+        if not diff:
```
kerry-meyer, line 209: This does not cover the case in which a subset of the existing breakout modes is specified in "want". If that occurs, "diff" will be empty and no action will occur. Instead, for this case, the breakout modes should be de-configured for all currently configured breakout interfaces except for the interface(s) specified in "want".

Please modify the logic here to handle this case.

plugins/module_utils/network/sonic/config/port_breakout/port_breakout.py
```diff
@@ -175,6 +176,56 @@ def _state_deleted(self, want, have, diff):
...
+            return commands, requests
+        commands = have
+        requests = self.get_delete_port_breakout_requests(commands, have)
+
+        if commands and len(requests) > 0:
+            send_requests(self._module, requests)
```
kerry-meyer, line 215: This will not generate the required output logs to show what has been deleted. Instead, "update_states" should be invoked to show the deleted breakout modes. 

Also, Instead of directly invoking "send_requests" here, it would be more efficient to just  append the required deletion requests to the list of requests to be executed on exit from the "_state_overridden" function.

plugins/module_utils/network/sonic/config/port_breakout/port_breakout.py
```diff
@@ -175,6 +176,56 @@ def _state_deleted(self, want, have, diff):
...
+        else:
+            commands = []
+
+        commands = diff
```
kerry-meyer, line 219: This will work if the modified deletion logic implemented to address the previous comment avoids deleting breakout modes that already match the requested new mode in "want". If the modified logic does a "delete all", then commands should be set equal to 'want' here.

tests/regression/roles/sonic_port_breakout/defaults/main.yml
```diff
@@ -26,10 +26,6 @@ tests:
     description: Configure breakout mode for ports
     state: merged
     input:
-      - name: 1/97
```
kerry-meyer, line 29: This eliminates the only "modify" cases for test_case_02.

Please restore these lines.

Instead of doing this, the "replaced" test  case can set the mode(s) back to their original value or to some other valid value. 

The "override" case can be kept as currently defined to test handling of a an "override" subset configuration, or, even better, could be supplemented to specify one interface breakout mode that is the same as the current one and one that differs from the current one.

kerry-meyer (changes requested):
- Functional changes in the "overridden" handling are needed and have been described in inline comments.

- Some minor changes have been requested in the regression test file.

- UT cases need to be added for "replaced" and "overridden state.

ArunSaravananBalachandran:
#291 has been raised to add support for 'replaced', 'overridden' states with addressing the review comments provided in this PR.
Hence closing this.

PR #207, opened 2023-04-01
Ut ntp

kannankvs:
##### SUMMARY
Unit testing code for NTP module added

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
NTP module Unit testing

PR #238, opened 2023-04-19
Added "replaced" and "overridden" states for the Interface module

divya-balasubramania:
##### SUMMARY
Added "replaced" and "overridden" states for the Interface module

##### COMPONENT NAME
Interface

##### OUTPUT
Interface execution log - 
[Interface_execution_overridden.zip](https://github.com/ansible-collections/dellemc.enterprise_sonic/files/11271691/Interface_execution_overridden.zip)

Interface module report - 
[Regression_Interface_Report.pdf] (https://github.com/ansible-collections/dellemc.enterprise_sonic/files/11271694/Regression_Interface_Report.pdf)

3 legacy test cases failed due to platform dependency. The same failures were observed without my changes as well.

Legacy run:-
            Interface module report -  [regression-legacy.pdf](https://github.com/ansible-collections/dellemc.enterprise_sonic/files/11271717/regression-legacy.pdf)

### DIFF
--- a/plugins/module_utils/network/sonic/config/interfaces/interfaces.py
+++ b/plugins/module_utils/network/sonic/config/interfaces/interfaces.py
@@ -175,7 +175,7 @@ class Interfaces(ConfigBase):
         """
         commands = self.filter_comands_to_change(diff, have)
         requests = self.get_delete_interface_requests(commands, have)
-        requests.extend(self.get_modify_interface_requests(commands, have))
+        requests.extend(self.get_modify_interface_requests(want, have))
         if commands and len(requests) > 0:
             commands = update_states(commands, "replaced")
         else:
@@ -194,7 +194,9 @@ class Interfaces(ConfigBase):
         """
         commands = []
         requests = []
-        if not diff:
+
+        diff1 = get_diff(have, want)
+        if not diff and not diff1:
             return commands, requests
 
         commands = have
@@ -203,12 +205,10 @@ class Interfaces(ConfigBase):
             send_requests(self._module, requests)
         else:
             commands = []
-        exist_interfaces_facts = self.get_interfaces_facts()
-        have_new = exist_interfaces_facts
         commands_over = get_diff(want, have)
-        requests = self.get_modify_interface_requests(commands_over, have_new)
+        requests = self.get_modify_interface_requests(diff, have)
 
-        if commands_over and len(requests) > 0:
+        if len(requests) > 0:
             commands = update_states(commands_over, "overridden")
 
         return commands, requests
--- a/plugins/module_utils/network/sonic/config/interfaces/interfaces.py
+++ b/plugins/module_utils/network/sonic/config/interfaces/interfaces.py
@@ -36,6 +36,7 @@ from ansible_collections.dellemc.enterprise_sonic.plugins.module_utils.network.s
 )
 from ansible_collections.dellemc.enterprise_sonic.plugins.module_utils.network.sonic.utils.utils import (
     get_diff,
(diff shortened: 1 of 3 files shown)


plugins/module_utils/network/sonic/config/interfaces/interfaces.py
```diff
@@ -192,18 +193,23 @@ def _state_overridden(self, want, have, diff):
...
-            commands.extend(commands_del)
-
-        commands_over = diff
-        requests.extend(self.get_modify_interface_requests(commands_over, have))
-        if commands_over and len(requests) > del_req_count:
-            commands_over = update_states(commands_over, "overridden")
-            commands.extend(commands_over)
+        requests = []
+        if not diff:
+            return commands, requests
```
mingjunzhang2019, line 198: This may not correct: if want is subset of have, diff=[], so the code does nothing. However in that case, want should override have for "overridden" state.

Similarly for the above "_state_replaced" function (could not add comment at it - line#76, so I add comment here). if the config for an interface in want is subset of those in have, the config for the interface in want should replace the config for same interface in have.

divya-balasubramania: corrected both functions.

plugins/module_utils/network/sonic/config/interfaces/interfaces.py
```diff
@@ -192,18 +193,23 @@ def _state_overridden(self, want, have, diff):
...
+        if not diff and not diff1:
+            return commands, requests
+
+        commands = have
+        requests = self.get_delete_all_interface(commands, have)
+        if commands and len(requests) > 0:
+            send_requests(self._module, requests)
+        else:
+            commands = []
+        commands_over = get_diff(want, have)
```
mingjunzhang2019, line 208: Does commands_over equal diff? If so, there is no use for commands_over.

plugins/module_utils/network/sonic/config/interfaces/interfaces.py
```diff
@@ -192,18 +193,23 @@ def _state_overridden(self, want, have, diff):
...
+        requests = self.get_modify_interface_requests(diff, have)
```
mingjunzhang2019, line 209: Line#202 to #207 deletes whole have. If want is subset of have, then diff is empty and requests is empty, so that there is nothing to apply.

plugins/module_utils/network/sonic/config/interfaces/interfaces.py
```diff
@@ -174,7 +175,7 @@ def _state_replaced(self, want, have, diff):
         """
         commands = self.filter_comands_to_change(diff, have)
         requests = self.get_delete_interface_requests(commands, have)
-        requests.extend(self.get_modify_interface_requests(commands, have))
+        requests.extend(self.get_modify_interface_requests(want, have))
```
mingjunzhang2019, line 178: Still if want is subset of have, then diff and commands are empty, so that there is nothing to be deleted and replaced. if the config for an interface in want is subset of those in have, the config for the interface in want should replace the config for same interface in have. But the code does nothing in such case.

kerry-meyer (changes requested):
Please address the comments posted by Mingjun for this PR.

(I am posting this comment to mark the PR with "changes requested".)

santhosh-kt:
Support for 'replaced' and 'overridden' states in `interface module` by addressing the review comments in this PR is being added as part of https://github.com/ansible-collections/dellemc.enterprise_sonic/pull/314.
Hence closing this.

PR #241, opened 2023-04-20
l3_interfaces - Add support for replaced and overridden states

santhosh-kt:
##### SUMMARY

Added "replaced" and "overridden" states for the l3_Interfaces module

##### GitHub Issues

| GitHub Issue # |
| -------------- |
| |

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

- sonic_l3_interfaces

##### OUTPUT

```paste below

```
##### ADDITIONAL INFORMATION

```paste below

```

##### Checklist:

- Regression report:

[regression-2023-11-29-13-59-53.zip](https://github.com/ansible-collections/dellemc.enterprise_sonic/files/13503569/regression-2023-11-29-13-59-53.zip)


tests/regression/roles/sonic_l3_interfaces/defaults/main.yml
```diff
@@ -234,6 +234,50 @@ tests:
               - address: 3051::1/16
               - address: 3052::1/16
   - name: test_case_11
+    description: Naming tests
```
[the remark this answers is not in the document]
santhosh-kt: Updated the test cases names. Sorry for delay reply. Notification is masked.

plugins/module_utils/network/sonic/config/l3_interfaces/l3_interfaces.py
```diff
@@ -164,16 +166,17 @@ def _state_overridden(self, want, have, diff):
...
-            delete_interfaces_requests = self.get_delete_l3_interfaces_requests(want, have)
-            ret_requests.extend(delete_interfaces_requests)
-            commands.extend(update_states(interfaces_to_delete, "deleted"))
+        new_want = self.update_object(want)
+        l3_interfaces_to_delete = self.get_interface_object_for_overridden(have)
+        diff = get_diff(l3_interfaces_to_delete, new_want, TEST_KEYS)
 
         if diff:
-            interfaces_to_create_requests = self.get_create_l3_interfaces_requests(diff, have, want)
+            delete_interfaces_requests = self.get_delete_all_requests(diff)
```
kerry-meyer, line 174: This will only delete configuration for interfaces specified in the executing playbook, but "overridden" state is defined as replacing all configuration for the affected resource module with only the configuration specified in the playbook.

For L3 interfaces, we could modify this to avoid deleting configuration for the management 0 interface, but all other interface configuration should be deleted if the playbook specifies a change for any interface.

kerry-meyer (changes requested):

kerry-meyer:
There are a couple of things to take care of before merging this PR:

- The current version of the "overridden" handling is not operating according the specification. (Please see the related comment for this posted inline for the "config" file.)
 - The posted "Regression test detailed log" output is for an earlier version of the test file. Please post the results for the sanity test with the current version of this resource module (after making the change requested above).

plugins/module_utils/network/sonic/config/l3_interfaces/l3_interfaces.py
santhosh-kt: Updated the requested changes.
Now Mangement0 is exempted from delete handling.
Also, instead of diff, all l3 interfaces(except eth0) will be removed during overridden.

kerry-meyer (changes requested):
Thanks for making the requested revisions and posting updated test results.

The proposed diffs and regression test results look good. I have requested a minor change to provide a better name for a function that is too generic (not descriptive of what it does) and only used for "replaced" handling.

Please add UT test cases for "replaced" and "overridden" cases. (One for each state is enough.) Enough new code has been added to decrease the code coverage for the l3_interfaces config file by 5% from the previous base, and it would be good to cover the new code with these additional test cases.

plugins/module_utils/network/sonic/config/l3_interfaces/l3_interfaces.py
```diff
@@ -143,19 +143,21 @@ def _state_replaced(self, want, have, diff):
         """
         ret_requests = list()
         commands = list()
-        l3_interfaces_to_delete = get_diff(have, want, TEST_KEYS)
-        obj = self.get_object(l3_interfaces_to_delete, want)
-        diff = get_diff(obj, want, TEST_KEYS)
+        new_want = self.update_object(want)
+        l3_interfaces_to_delete = self.get_object(have, want)
```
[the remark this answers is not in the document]
santhosh-kt: Updated the description name.
Also updated the UT cases for overridden and replaced.

kerry-meyer:
Thanks for pointing out that you have already addressed my comments regarding UT test cases. I have re-checked the code change set and test results. Everything looks good except that regression test results need to be posted for the current version of the change set.

Please post updated regression test results and I expect that this PR can be approved and merged.

santhosh-kt:
@kerry-meyer : Updated the PR with latest regression results details.

kerry-meyer (commented):
Thank you for posting the regression results for the current version of the code submitted in this PR.

The "replaced" handling implemented here causes deletion of the address list for an AF (IPv6 or IPv4) on an interface when that AF is not present in the playbook at all if the playbook specifies an address or attribute for the "other" AF. An example of this behavior can be seen in the posted results for test_case_12. This is not the behavior we have specified for "replaced" state. It should be possible to replace the configuration for one AF of an interface without affecting the other AF.

If the current behavior is a conscious choice, please explain why it is implemented this way. Otherwise, it would be better to fix this before we merge this PR.

plugins/module_utils/network/sonic/config/l3_interfaces/l3_interfaces.py
```diff
@@ -164,16 +166,17 @@ def _state_overridden(self, want, have, diff):
...
         ret_requests = list()
         commands = list()
-        interfaces_to_delete = get_diff(have, want, TEST_KEYS)
-        if interfaces_to_delete:
-            delete_interfaces_requests = self.get_delete_l3_interfaces_requests(want, have)
-            ret_requests.extend(delete_interfaces_requests)
-            commands.extend(update_states(interfaces_to_delete, "deleted"))
+        new_want = self.update_object(want)
+        get_override_interfaces = self.get_interface_object_for_overridden(have)
+        diff = get_diff(get_override_interfaces, new_want, TEST_KEYS)
```
kerry-meyer, line 171: Coverage is needed here for the case of an "overridden" task that adds more addresses to an existing address list.

kerry-meyer:
> Thank you for posting the regression results for the current version of the code submitted in this PR.
> 
> The "replaced" handling implemented here causes deletion of the address list for an AF (IPv6 or IPv4) on an interface when that AF is not present in the playbook at all if the playbook specifies an address or attribute for the "other" AF. An example of this behavior can be seen in the posted results for test_case_12. This is not the behavior we have specified for "replaced" state. It should be possible to replace the configuration for one AF of an interface without affecting the other AF.
> 
> If the current behavior is a conscious choice, please explain why it is implemented this way. Otherwise, it would be better to fix this before we merge this PR.

Regarding the comment above that is quoted here:

Santhosh has pointed out that the Ansible L3 Interfaces "replaced" implementation for at least one other networking vendor does the replacement at the entire interface level if any attribute is modified. I re-checked an Ansible documentation source that we had previously referred to, and it has been modified to match this method for "replaced" handling for L3 interface configuration. Because of these considerations, I am willing to approve the "replaced" handling currently proposed in this PR.

(The "overridden" handling, as noted in my most recent inline comment, does still need to be examined and probably modified. Santhosh and I have discussed the "overridden" issue and my understanding is that he will post an update about this after doing some additional testing.)

plugins/module_utils/network/sonic/config/l3_interfaces/l3_interfaces.py
```diff
@@ -143,19 +143,21 @@ def _state_replaced(self, want, have, diff):
         """
         ret_requests = list()
         commands = list()
-        l3_interfaces_to_delete = get_diff(have, want, TEST_KEYS)
-        obj = self.get_object(l3_interfaces_to_delete, want)
-        diff = get_diff(obj, want, TEST_KEYS)
+        new_want = self.update_object(want)
+        get_replace_interfaces_list = self.get_interface_object_for_replaced(have, want)
+
```
kerry-meyer, line 148: The deletion portion of this block has the following problems that need to be fixed (in the section from here through the current line 154):

- No deletion requests should be sent to the device for interfaces that currently have no L3 configuration. The test output indicates that a "delete" request is sent to interface Etherneret1/6 for a "replaced" configuration that touches that interface even though the "before" L3 configuration for this interface is empty.
- In the reported output for "deleted" commands, the deleted values displayed should be the values that were configured at the start of execution of the task (the values that were actually deleted). This corresponds to the subset of "have" specified in "get_replace_interfaces_list", not to "want" as currently coded in line 154.

tests/regression/roles/sonic_l3_interfaces/defaults/main.yml
kerry-meyer, line 1: Please add merged configuration to some more interfaces (at least 1 more interface) preceding the "replaced" and "overridden" test cases to enable verification that:

- A "replaced" task does not affect interfaces that are not specified in the task.
- An "overridden" task deletes configuration for all interfaces if it is making any changes.

kerry-meyer (changes requested):
Thanks for reminding me about the pending updates for this module. The 'overridden' handling looks good now.

I have posted a couple of small additional change requests. It looks like this PR can be approved and merged after these additional changes are made.

tests/regression/roles/sonic_l3_interfaces/defaults/main.yml
santhosh-kt: Added new test case(test_case_11). 
That helped to verify both the cases.

plugins/module_utils/network/sonic/config/l3_interfaces/l3_interfaces.py
kerry-meyer: After the most recent push, the revised behavior looks fine. 

Thanks for making this additional revision.

Resolved.

kerry-meyer (approved):
The current proposed changes and posted test results all look good.

Thank you for providing this additional functionality.

Approved.

PR #273, opened 2023-06-02
Resolved regression failures

stalabi1:
##### SUMMARY

Resolved regression failures

##### ISSUE TYPE

- Test Pull Request

##### COMPONENT NAME

sonic_l3_interfaces
sonic_vrfs
sonic_logging

##### OUTPUT
[regression-2023-06-02-15-16-19.html.pdf](https://github.com/ansible-collections/dellemc.enterprise_sonic/files/11639479/regression-2023-06-02-15-16-19.html.pdf)

### DIFF
--- a/tests/regression/roles/sonic_vrfs/tasks/main.yml
+++ b/tests/regression/roles/sonic_vrfs/tasks/main.yml
@@ -10,9 +10,10 @@
   include_tasks: tasks_template.yaml
   loop: "{{ tests }}"
 
-- name: Cleanup tests
-  include_tasks: cleanup_tests.yaml
-
-# - name: Display all variables/facts known for a host
-#   debug:
-#     var: hostvars[inventory_hostname].ansible_facts.test_reports
+- name: "Delete interfaces"
+  vars:
+    ansible_connection: network_cli
+  sonic_config:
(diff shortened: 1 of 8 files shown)


tests/regression/roles/sonic_vrfs/tasks/main.yml
kerry-meyer, line 1: Although the current "cleanup_tests.yaml" is not useful now, it would be good to modify it so it deletes the interfaces that were created for this test and add a "Cleanup" task in this file to execute the revised "cleanup_tests.yaml" file (or whatever it's new name becomes; see "Note" below) after running the test so the interfaces created for this test are removed.

Note:  The modified "cleanup_tests.yaml" file should probably be renamed because it would not really be a "test" of VRF functionality if it just deletes interfaces.

kerry-meyer (changes requested):
These changes look like they will fix the observed failure.

I am concerned, though, that the created interfaces are apparently being left configured. I have posted a comment with more details regarding this consideration.

kerry-meyer (approved):
Thank you for adding the post-test deletion of the logical interfaces created for this test.

Approved.

PR #408, opened 2024-06-26
Update AAA regression test cases for 2.5.0

stalabi1:
##### SUMMARY

I commented out AAA regression test case 2 because such configuration will cause a lockout on the switch during regression execution.

##### ISSUE TYPE

- Test Pull Request

##### COMPONENT NAME

sonic_aaa

##### OUTPUT

[regression-2024-06-26-12-39-57.html.pdf](https://github.com/user-attachments/files/15994551/regression-2024-06-26-12-39-57.html.pdf)

### DIFF
--- a/tests/regression/roles/sonic_aaa/defaults/main.yml
+++ b/tests/regression/roles/sonic_aaa/defaults/main.yml
@@ -12,13 +12,14 @@ tests:
           group: tacacs+
           local: true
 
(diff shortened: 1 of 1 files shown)


kerry-meyer (approved):
Thank you for your fast action on disabling this connection-breaking regression test to allow running of the full automated regression test suite.

Approved.

PR #441, opened 2024-08-23
Enable system auditd and ip loadshare hash downward compatibility

kerry-meyer:
##### SUMMARY

Modify the enterprise_sonic httapi plugin "edit_config" wrapper for invocation of the Ansible "send" utility to avoid playbook-aborting exceptions due to the unconditional "GET" REST API invocation done within the "system" resource module for the auditd and "ip loadshare hash" configuration. Modify this wrapper so that, when the new "suppress notification of exception" (suppr_ntf_excp) input flag is True, it returns a valid, but empty "response" for any GET request for which a "resource not found" (code '404') ConnectionError exception occurs instead of aborting the currently executing playbook with the ConnectionError exception. Because the auditd and "ip loadshare hash" configurations are only available in SONiC 4.4+, invoking these GETs on older versions of SONIC causes a ConnectionError exception to be raised in the Ansible "send" method of the HttpApi "connection"  object when a "resource not found" error is returned for the GET.

Because the "system" resource module facts collection function unconditionally performs the auditd and "ip loadshare hash algorithm" GET operations regardless of the SONiC version on the target device, this change is needed to enable downward compatibility of the "system" resource module for SONiC versions before 4.4.0. 

##### GitHub Issues

| GitHub Issue # |
| -------------- |
| |
N/A

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

sonic_system
##### OUTPUT

Regression test on a Z9664f switch (supporting ip loadshare hash algorithm configuration) with a recent 4.x_share image:

[system_downward_with_loadshare_latest_regression-2024-08-22-21-23-06.pdf](https://github.com/user-attachments/files/16724939/system_downward_with_loadshare_latest_regression-2024-08-22-21-23-06.pdf)

Regression test (conditional filter) with ip loadshare hash testing commented out on an s5296f switch, recent 4.x_share image:

[system_downward_conditional_httapi_filter_regression-2024-08-27-18-49-10.pdf](https://github.com/user-attachments/files/16776939/system_downward_conditional_httapi_filter_regression-2024-08-27-18-49-10.pdf)

Regression test on modules requiring fixes for empty GET response handling:

Note: The PoE test is not supported on the available test platforms, so only the PoE Unit Tests have been run.

[system_downward_empty_config_GET_fixes_regression-2024-08-29-04-49-32.zip](https://github.com/user-attachments/files/16796408/system_downward_empty_config_GET_fixes_regression-2024-08-29-04-49-32.zip)

##### ADDITIONAL INFORMATION

System "gather facts" on a SONiC 4.2.1 image without the fix:

[gather_system_SONiC_4_2_1_failure.log](https://github.com/user-attachments/files/16724764/gather_system_SONiC_4_2_1_failure.log)

System "gather facts" on a SONiC 4.2.1 image with the conditionally enabled GET exception filter:

[sonic_api_get_httpapi_filter_SONiC_4_2_1_PR_mod.log](https://github.com/user-attachments/files/16776781/sonic_api_get_httpapi_filter_SONiC_4_2_1_PR_mod.log)

sonic_api "get" on a SONiC 4.x_share image with the conditionally enabled GET exception filter:

[sonic_api_get_httpapi_filter_SONiC_4_2_1_PR_mod.log](https://github.com/user-attachments/files/16776745/sonic_api_get_httpapi_filter_SONiC_4_2_1_PR_mod.log)

sonic_cli "show" on a SONiC 4.x_share image with the conditionally enabled GET exception filter:

[sonic_cli_show_httpapi_filter_SONiC_4_2_1_PR_mod.log](https://github.com/user-attachments/files/16776798/sonic_cli_show_httpapi_filter_SONiC_4_2_1_PR_mod.log)

PoE "gather facts" on a platform that doesn't support PoE (causing the empty config case that needs to be verified to confirm that empty configuration is returned and no exception occurs):

[poe_gather_facts_handle_empty.log](https://github.com/user-attachments/files/16796335/poe_gather_facts_handle_empty.log)

##### Checklist:

See output above: I ran a playbook to gather "system" facts against a SONiC 4.2.1 image without the fix to demonstrate the failure and with the fix to demonstrate the effectiveness of the fix. I also ran the full "system" regression test on a z9664 switch supporting loadshare, and also the proposed "stubbed" system regression test with ip loadshare hash testing disabled on an s5296f.

### DIFF
--- a/changelogs/fragments/441-system-downward-compatibility-fix.yaml
+++ b/changelogs/fragments/441-system-downward-compatibility-fix.yaml
@@ -1,3 +1,3 @@
 ---
 bugfixes:
-  - system - Catch the ConnectionError exception caused by unconditional fetching of auditd and ip loadshare hash algorithm configuration fetches and return empty configuration instead of allowing the uncaught exception to abort all "system" operations on SONiC images older than version 4.4.0 (https://github.com/ansible-collections/dellemc.enterprise_sonic/pull/434).
+  - sonic_system - Catch the ConnectionError exception caused by unconditional fetching of auditd and ip loadshare hash algorithm configuration, and return empty configuration instead of allowing the uncaught exception to abort all "system" operations on SONiC images older than version 4.4.0 (https://github.com/ansible-collections/dellemc.enterprise_sonic/pull/441).
--- a/changelogs/fragments/441-system-downward-compatibility-fix.yaml
+++ b/changelogs/fragments/441-system-downward-compatibility-fix.yaml
@@ -1,0 +1,3 @@
+---
+bugfixes:
+  - system - Catch the ConnectionError exception caused by unconditional fetching of auditd and ip loadshare hash algorithm configuration fetches and return empty configuration instead of allowing the uncaught exception to abort all "system" operations on SONiC images older than version 4.4.0 (https://github.com/ansible-collections/dellemc.enterprise_sonic/pull/434).
--- a/plugins/module_utils/network/sonic/sonic.py
+++ b/plugins/module_utils/network/sonic/sonic.py
@@ -138,6 +138,7 @@ def edit_config(module, commands, skip_code=None):
     # End
     return connection.edit_config(commands)
 
+
 def edit_config_catch(module, commands, skip_code=None):
     connection = get_connection(module)
 
@@ -154,6 +155,7 @@ def edit_config_catch(module, commands, skip_code=None):
     except ConnectionError:
         response = [[{}, {}]]
     return response
+
 
 def edit_config_reboot(module, commands, skip_code=None):
     connection = get_connection(module)
--- a/plugins/module_utils/network/sonic/sonic.py
+++ b/plugins/module_utils/network/sonic/sonic.py
@@ -125,7 +125,7 @@ def run_commands(module, commands, check_rc=True):
         module.fail_json(msg=to_text(exc))
 
 
-def edit_config(module, commands, skip_code=None):
+def edit_config(module, commands, skip_code=None, suppr_ntf_excp=True):
     connection = get_connection(module)
 
     # Start: This is to convert interface name from Eth1/1 to Eth1%2f1
@@ -136,7 +136,11 @@ def edit_config(module, commands, skip_code=None):
             if url:
                 request["path"] = update_url(url)
     # End
-    return connection.edit_config(commands)
+    if suppr_ntf_excp:
+        # Default: not used for cliconf
+        return connection.edit_config(commands)
+    else:
+        return connection.edit_config(commands, suppr_ntf_excp)
 
 
 def edit_config_reboot(module, commands, skip_code=None):
--- a/plugins/module_utils/network/sonic/sonic.py
+++ b/plugins/module_utils/network/sonic/sonic.py
@@ -138,6 +138,22 @@ def edit_config(module, commands, skip_code=None):
     # End
     return connection.edit_config(commands)
 
+def edit_config_catch(module, commands, skip_code=None):
+    connection = get_connection(module)
+
+    # Start: This is to convert interface name from Eth1/1 to Eth1%2f1
+    for request in commands:
+        # This check is to differenciate between requests and commands
+        if isinstance(request, dict):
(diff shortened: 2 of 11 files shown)


changelogs/fragments/441-system-downward-compatibility-fix.yaml
```diff
@@ -0,0 +1,3 @@
+---
+bugfixes:
+  - system - Catch the ConnectionError exception caused by unconditional fetching of auditd and ip loadshare hash algorithm configuration fetches and return empty configuration instead of allowing the uncaught exception to abort all "system" operations on SONiC images older than version 4.4.0 (https://github.com/ansible-collections/dellemc.enterprise_sonic/pull/434).
```
ArunSaravananBalachandran, line 3:
```suggestion
  - sonic_system - Catch the ConnectionError exception caused by unconditional fetching of auditd and ip loadshare hash algorithm configuration and return empty configuration instead of allowing the uncaught exception to abort all "system" operations on SONiC images older than version 4.4.0 (https://github.com/ansible-collections/dellemc.enterprise_sonic/pull/434).
```

ArunSaravananBalachandran (changes requested):

changelogs/fragments/441-system-downward-compatibility-fix.yaml
kerry-meyer: Thanks for catching that grammar error in the description and the missing "sonic_" prefix on the component name.

I will fix this, but am not committing it "as is" because I see that I also forgot to change the copied PR number from "434" to "441".

kerry-meyer: Fixed.

plugins/module_utils/network/sonic/sonic.py
```diff
@@ -139,6 +139,24 @@ def edit_config(module, commands, skip_code=None):
...
+            if url:
+                request["path"] = update_url(url)
+    # End
+    try:
+        response = connection.edit_config(commands)
+    except ConnectionError:
+        response = [[{}, {}]]
+    return response
+
+
```
mingjunzhang2019, line 159: I Suppose this new function will be used by "GET" request only. It is proper to return "[[{}, {}]]" if target system return error 404 (resource not found). How about other errors returned from target system? or not expect other errors from target system for "GET" request?

kerry-meyer: Yes. I think this handling should be conditional on a 404 error, maybe only when the request is a GET. I am also thinking that the fix needs to be lower in the call stack to enable handling for an individual request in the list of requests. (It is currently handing an exception occurring anywhere during the iterative loop through the requests at the "connection type" edit_config function.)

I will do some rework on the fix to address these concerns.

mingjunzhang2019 (approved):
"edit_config" is great!

tests/regression/roles/sonic_system/defaults/main.yml
```diff
@@ -12,7 +12,8 @@ tests:
         ipv4: false
         ipv6: false
       auto_breakout: ENABLE
-      load_share_hash_algo: JENKINS_HASH_HI
+      # Use only on switch models that support this.
+      #load_share_hash_algo: JENKINS_HASH_HI
```
mingjunzhang2019, line 16: "load_share_hash_algo"'s are commented out. Are those intended?

plugins/module_utils/network/sonic/sonic.py
kerry-meyer: I have modified the change set to use the suggested filtering criteria plus an additional flag specified by the application to enable/disable the filtering.

tests/regression/roles/sonic_system/defaults/main.yml
kerry-meyer: Yes. This test is only supported on a Z9664f switch. It is commented out instead of deleted so that we can still run it on a Z9664f by uncommenting the indicated lines of the test.

mingjunzhang2019 (approved):

ArunSaravananBalachandran (approved):

PR #490, opened 2024-12-04
Add FBS classifiers resource module

stalabi1:
##### SUMMARY

I added the FBS classifiers resource module.
Please see [model PR](https://github.com/ansible-network/resource_module_models/pull/284).
##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

sonic_fbs_classifiers

##### OUTPUT

[regression-2025-01-16-15-08-12.html.pdf](https://github.com/user-attachments/files/18446296/regression-2025-01-16-15-08-12.html.pdf)
[facts_gathering.log](https://github.com/user-attachments/files/18015393/facts_gathering.log)
[diff_output.log](https://github.com/user-attachments/files/18015396/diff_output.log)
[check_mode_output.log](https://github.com/user-attachments/files/18015398/check_mode_output.log)
[ansible_lint_output.txt](https://github.com/user-attachments/files/18303742/ansible_lint_output.txt)

##### Checklist:


plugins/module_utils/network/sonic/config/fbs_classifiers/fbs_classifiers.py
```diff
@@ -0,0 +1,602 @@
...
+            match_type = classifier.get('match_type')
+            cfg_class_description = cfg_classifier.get('class_description')
+            cfg_match_type = cfg_classifier.get('match_type')
+
+            if ((class_description and cfg_class_description and class_description != cfg_class_description) or
+                    (match_type and cfg_match_type and match_type != cfg_match_type)):
+                requests.append(self.get_delete_classifiers_request(class_name))
+                config_list.append(cfg_classifier)
+                break
+
```
kerry-meyer, line 558: For the remaining parts of the classifier, if the playbook does not specify the corresponding part (e.g. match_acl) but that part is currently configured, it would be more consistent with our current "replaced" state specification to delete that part from the configuration.

The same consideration applies to the class_description: If the "replaced" playbook does not specify it and it was previously configured, it should be deleted.

plugins/modules/sonic_fbs_classifiers.py
```diff
@@ -0,0 +1,472 @@
...
+              vlanid:
+                description:
+                  - VLAN ID, range 1-4094
+                type: int
+          transport:
+            description:
+              - Transport field configuration
+            type: dict
+            suboptions:
+              source_port:
```
kerry-meyer, line 153: Please include in the port descriptions a description of the range delimiter used for this resource module.

It would be more consistent with other modules in our collection to use '-' as the range delimiter, but it's okay to use ".." instead if it is described here. Even if the '-' notation is used, it would be worthwhile to state that in the description.

kerry-meyer (changes requested):
The structure and most details of the proposed code look good. The corresponding test results also look good.

I am requesting a few changes, mostly minor.

I would like to re-check the test results after the changes to verify the requested change in "replaced" state behavior.

plugins/modules/sonic_fbs_classifiers.py
```diff
@@ -0,0 +1,484 @@
...
+#     tcp-flags fin psh ack
+#   Referenced in flows:
+
+
+# Using Replaced
+#
+# Before state:
+# -------------
+#
+# sonic# show class-map
```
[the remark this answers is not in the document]
kerry-meyer: This comment was already "Pending" when we discussed that. Thanks for doing the update. I will mark this "Issue" as "Resolved".

plugins/module_utils/network/sonic/config/fbs_classifiers/fbs_classifiers.py
kerry-meyer: In considering our current "replaced state handling" specification, I think that the argspec organization for this resource module justifies an exception to the current rules. The current replacement granularity is consistent with the intent of the specification. Although a list of lists is covered in the specification, which specifies the replacement granularity for that type of argspec as a second level list item, a dictionary of dictionaries (which is the case here) is not covered by the specification, although it is a similar situation and should be included in the specification. The granularity implemented here is reasonable. 

So the only remaining portion of this "Issue" is that still needs to be addressed is deletion of a classifier if one of the top level non-key options, "match_type" or "description" is included in the playbook (even with a value the same as the current value) and the other top level non-key option is not included in the playbook.

kerry-meyer: With the latest changes, the case described in the last paragraph of my previous comment will not result in the implicit deletion of the classifier for the following case:

match_acl and/or match_hdr are not empty and one one of the following is true:
- There is a change in class_description or match_type
- One of the top level non-key options, "match_type" or "class_description" is included in the playbook (even with a value the same as the current value) and the other top level non-key option is not included in the playbook.

kerry-meyer: With the latest changes, the handling looks fine.

kerry-meyer (approved):
All proposed code in the current version of the change set and the corresponding test results look good.

Approved.

PR #505, opened 2025-01-27
[SFT-805] Ansible resource module changes for Concurrent session limit

balasubramaniam-k:
**SUMMARY**
Resource module changes for Concurrent session limit funtionality

**ISSUE TYPE**
- Feature Pull Request

**COMPONENT NAME**
sonic_system

**RELATED PR**

**OUTPUT**
[regression-2025-01-13-17-42-51.html.pdf](https://github.com/user-attachments/files/18554019/regression-2025-01-13-17-42-51.html.pdf)
[regression_system_diff_mode.log](https://github.com/user-attachments/files/18554023/regression_system_diff_mode.log)
[system_check_mode.log](https://github.com/user-attachments/files/18554024/system_check_mode.log)

**Checklist:**


balasubramaniam-k:
> The change set looks good. For the check mode log, could you append a delete case where concurrent_session_limit is configured in the before then delete that concurrent_session_limit in check mode. I see a case where concurrent_session_limit 5 is deleted but it wasn't previously configured.

@stalabi1, thanks for the review. Just a correction, concurrent session limit is not deleted since it's not part of before as shown in the previous log.
I have attached the check mode log with concurrent session limit in before and it being deleted.

awhaley-dell (approved):
The code and tests look good to me. Thanks for providing all of the documentation of your testing as well.

plugins/modules/sonic_system.py
```diff
@@ -269,20 +285,20 @@
...
-# ip anycast-address enable
 # ipv6 anycast-address enable
 # ip load-share hash algorithm JENKINS_HASH_HI
+# login concurrent-session limit 5
 
 - name: Override system configuration.
   sonic_system:
     config:
-      hostname: sonic
+      hostname: SONIC
```
kerry-meyer, line 295: What is the reason for this line change? I'm assuming it is here to show application of a non-default hostname.

The resulting non-default "hostname" option should be displayed in the "after" instance of "show running-configuration". (Actually, even the **default** value of "sonic" is displayed with "show running-configuration", so either way it should always have been here, although I realize that it was missing from the base example.)

kerry-meyer (changes requested):
Thank you for providing the code changes to add this new functionality.

I am posting a couple of minor change requests for Examples in the "modules" file which otherwise looks fine.

I would still like to do a quick check on the other files before Approving and merging this PR. If the requested changes are made by tomorrow in Santa Clara, I will try to get the remaining part of the review done, including merging of the PR if no other issues are found.

plugins/modules/sonic_system.py
```diff
@@ -216,20 +227,25 @@
...
 
 - name: Replace system configuration.
   sonic_system:
     config:
-      hostname: sonic
-      interface_naming: standard
+      hostname: SONIC
+      anycast_address:
+        ipv6: true
+      concurrent_session_limit: 5
```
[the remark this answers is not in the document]
balasubramaniam-k: Agreed. Code changes are made.

plugins/modules/sonic_system.py
kerry-meyer: Thanks for explaining. This was a worthwhile correction to make.

Although the "after state" does show that the host **prompt** has changed to upper case, it doesn't show this configuration line

```
# !
# hostname SONIC       <<<
```

 in the "show running-configuration" sample output.

```
# !
# ip anycast-mac-address bb:aa:cc:dd:ee:ff
# ip anycast-address enable
# interface-naming standard
# ip load-share hash algorithm CRC_XOR
# login concurrent-session limit 4
```


 It is also missing from the "before state" for this example. It is there in the "before state" for the next example, but missing in both the "before state" (lower case) and "after state" (upper case) for the "override" example. This is a minor point, but please include this configuration line in the displayed "show running-configuration" output for both "before" and "after".

plugins/modules/sonic_system.py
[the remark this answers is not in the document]
kerry-meyer: The update looks fine. Thank you.

Resolved.

plugins/modules/sonic_system.py
balasubramaniam-k: As discussed, displaying default configuration in running configuration will be masked in the near future and hence it's not necessary to update "before" in this instance. "after" is updated appropriately.

kerry-meyer (approved):
The current proposed set of code changes and corresponding test results look good.

Approved.

Issue #502, opened 2025-01-15
[BUG]: sonic_stp mstp adding a vlan to a an mst instance removes all previously defined vlans from that instance 


TafkaMax:
### Bug Description

When adding a new vlan to an MST instance the previously defined vlans are removed from the mst instance. 

```
- name: Merge STP configuration for Dell Sonic
  dellemc.enterprise_sonic.sonic_stp:
    config: '{{ sonic__stp }}'
    state: replaced
```

The output of the sonic_stp variable is below.

```
ok: [redacted] => {
    "sonic__stp": {
        "global": {
            "bpdu_filter": true,
            "bridge_priority": 4096,
            "enabled_protocol": "mst",
            "loop_guard": true
        },
        "mstp": {
            "mst_instances": [
                {
                    "bridge_priority": 4096,
                    "mst_id": 0
                },
                {
                    "bridge_priority": 4096,
                    "mst_id": 1,
                    "vlans": [
                        "20",
                        "90"
                    ]
                },
                {
                    "bridge_priority": 4096,
                    "mst_id": 2,
                    "vlans": [
                        "14",
                        "22",
                        "70",
                        "72",
                        "91"
                    ]
                },
                {
                    "bridge_priority": 4096,
                    "mst_id": 3,
                    "vlans": [
                        "21",
                        "23",
                        "69",
                        "71",
                        "78",
                        "92"
                    ]
                },
                {
                    "bridge_priority": 4096,
                    "mst_id": 15,
                    "vlans": [
                        "3000"
                    ]
                }
            ],
            "mst_name": "redacted",
            "revision": 12
        }
    }
}
```

### Product Name

Dell Enterprise Sonic

### Component or Module Name

httpapi

### DellEMC Enterprise SONiC Ansible Collection Version

latest git hash

### SONiC Software Version

4.4.1-Enterprise_Standard

### Configuration

```
CONFIG_FILE() = /home/tansper/ansible/ansible.cfg
DEFAULT_HOST_LIST(/home/tansper/ansible/ansible.cfg) = ['/home/tansper/ansible/hosts']
DEFAULT_LOOKUP_PLUGIN_PATH(/home/tansper/ansible/ansible.cfg) = ['/home/tansper/ansible/plugins/lookup']
DEFAULT_VAULT_PASSWORD_FILE(/home/tansper/ansible/ansible.cfg) = /home/tansper/ansible/scripts/vault_password.sh
[...]

PR #511, opened 2025-02-18
STP bugfix: Add handling for removal of empty data structures for merge state

stalabi1:
##### SUMMARY

In the case of configuring vlans for an mst instance, there is some special handling to ensure that a request for an already configured vlan is not generated. The handling involves popping the vlan list from the mst instance if empty. There was no logic to check if the encapsulating data structures were empty and needed popping so the commands list would be incorrect for certain cases.  I added such logic.

##### GitHub Issues

| GitHub Issue #502  |
| -------------- |

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

sonic_stp

##### OUTPUT

Regression report:
[regression-2025-02-25-10-49-54.html.pdf](https://github.com/user-attachments/files/18970849/regression-2025-02-25-10-49-54.html.pdf)

mst_instances interfaces error handling:
[error_handling.txt](https://github.com/user-attachments/files/18914314/error_handling.txt)

##### Checklist:

### DIFF
--- a/plugins/module_utils/network/sonic/config/stp/stp.py
+++ b/plugins/module_utils/network/sonic/config/stp/stp.py
@@ -224,7 +224,6 @@ class Stp(ConfigBase):
 
         if replaced_config:
             commands.extend(update_states(replaced_config, "deleted"))
-            mod_commands = want
 
             for command in commands:
                 self.transform_config_for_diff_check(command)
@@ -232,6 +231,7 @@ class Stp(ConfigBase):
             new_config = get_new_config(commands, tmp_have, TEST_KEYS_generate_config)
             new_config = self.post_process_generated_config(new_config)
             tmp_have = new_config
+            mod_commands = get_diff(want, tmp_have, TEST_KEYS)
 
         else:
             mod_commands = diff


kerry-meyer (changes requested):
The proposed code changes and corresponding posted test results look good.

Since the case of adding only vlans to an MST instance was found to be a problem case, it would be best to add a new test item to the regression test suite to check and verify initial and idempotent handling for this case. This would help to prevent any future related regression problems.

plugins/module_utils/network/sonic/config/stp/stp.py
```diff
@@ -220,16 +219,25 @@ def _state_replaced(self, want, have, diff):
...
         if replaced_config:
             commands.extend(update_states(replaced_config, "deleted"))
             mod_commands = want
+
+            for command in commands:
+                self.transform_config_for_diff_check(command)
+            self.transform_config_for_diff_check(tmp_have)
+            new_config = get_new_config(commands, tmp_have, TEST_KEYS_generate_config)
+            new_config = self.post_process_generated_config(new_config)
+            tmp_have = new_config
```
kerry-meyer, line 234: Instead of using all of the commands in "want" for mod_commands for this case, it would be more efficient to set "mod_commands = get_diff(want, tmp_have, TEST_KEYS)"

plugins/module_utils/network/sonic/config/stp/stp.py
kerry-meyer, line 1: Does the fix applied for this PR in the "functional" config handling also need to be done for "check" mode handling?

kerry-meyer:
> The proposed code changes and corresponding posted test results look good.
> 
> Since the case of adding only vlans to an MST instance was found to be a problem case, it would be best to add a new test item to the regression test suite to check and verify initial and idempotent handling for this case. This would help to prevent any future related regression problems.

The suggestion quoted here has been implemented as requested. This "Issue" is "Resolved".

plugins/module_utils/network/sonic/config/stp/stp.py
kerry-meyer: Thanks for adding this additional enhancement.

Resolved.

plugins/module_utils/network/sonic/config/stp/stp.py
kerry-meyer: After walking through the relevant portions of the code with you, I agree that this case is correctly covered by the existing code. This is because the "check mode" deletion phase (for "replaced" and "overridden" states) occurs before and separately from the "replaced/merged" handling. This should work as intended to yield the correct result when all vlans in an MST instance are deleted during "check mode" processing.

Resolved.

PR #521, opened 2025-03-04
Add mirroring resource module

stalabi1:
##### SUMMARY

I am adding the sonic_mirroring resource module on behalf of Mingjun. I addressed the PR comments that were posted in #415.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

sonic_mirroring

##### OUTPUT

[regression-2025-04-21-12-00-28.html.pdf](https://github.com/user-attachments/files/19839530/regression-2025-04-21-12-00-28.html.pdf)

##### Checklist:

- [] I have maintained at least 90% code coverage

### DIFF
--- a/plugins/module_utils/network/sonic/config/mirroring/mirroring.py
+++ b/plugins/module_utils/network/sonic/config/mirroring/mirroring.py
@@ -40,7 +40,7 @@ from ansible_collections.dellemc.enterprise_sonic.plugins.module_utils.network.s
 
 PATCH = 'patch'
 DELETE = 'delete'
-URL = 'data/sonic-mirror-session:sonic-mirror-session/MIRROR_SESSION/MIRROR_SESSION_LIST'
+URL = 'data/openconfig-mirror-ext:mirror/sessions'
 TEST_KEYS = [
     {'span': {'name': ''}},
     {'erspan': {'name': ''}},
@@ -187,7 +187,7 @@ class Mirroring(ConfigBase):
         command = diff
         requests = self.get_modify_mirroring_requests(command)
         if command and len(requests) > 0:
-            commands = update_states([command], "merged")
+            commands = update_states([command], 'merged')
         else:
             commands = []
         return commands, requests
@@ -199,7 +199,6 @@ class Mirroring(ConfigBase):
         :param have: the current configuration as a dictionary
         :rtype: A list
         :returns: the commands necessary to remove the specified playbook options from the current configuration
-                  of the provided objects
         """
         # if want is none, then delete all the mirroring except admin
         commands = []
@@ -214,7 +213,7 @@ class Mirroring(ConfigBase):
         requests = self.get_delete_mirroring_requests(command, delete_all)
 
         if command and len(requests) > 0:
-            commands = update_states([command], "deleted")
+            commands = update_states([command], 'deleted')
 
         return commands, requests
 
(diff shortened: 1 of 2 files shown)


plugins/module_utils/network/sonic/config/mirroring/mirroring.py
```diff
@@ -0,0 +1,461 @@
...
+)
+from ansible_collections.dellemc.enterprise_sonic.plugins.module_utils.network.sonic.utils.formatted_diff_utils import (
+    __DELETE_CONFIG_IF_NO_SUBCONFIG,
+    get_new_config,
+    get_formatted_config_diff
+)
+
+PATCH = 'patch'
+DELETE = 'delete'
+URL = 'data/sonic-mirror-session:sonic-mirror-session/MIRROR_SESSION/MIRROR_SESSION_LIST'
```
kerry-meyer, line 43: The "facts" file for this resource module is using openconfig. Unless there is a good reason not to, this file should also use it. The main difference in code here would be the conversion of the argspec "_" to the Opencofing "-" for applicable options. (The REST API structure below the top level looks the same, but please verify that and correct the path specifications as needed.)

```suggestion
URL = 'data/openconfig-mirror-ext:mirror/sessions/session'
```

kerry-meyer (changes requested):
I have reviewed all files except for the test files.

I am posting my current "pending" comments so they can be addressed before I complete the review for the other files. (Those files will be affected by the changes requested here, but I will proceed with checking of any generic test file code that is applicable regardless of the pending changes in parallel with the coding of changes to address this initial set of comments.)

plugins/module_utils/network/sonic/config/mirroring/mirroring.py
```diff
@@ -0,0 +1,479 @@
...
+        if span and erspan:
+            for ms in span:
+                name = ms['name']
+                in_erspan = next((ems for ems in erspan if name == ems['name']), None)
+                if in_erspan:
+                    err_msg = "Names of SPAN and ERSPAN mirror sessions should not be duplicated."
+                    self._module.fail_json(msg=err_msg, code=400)
+
+    def preprocess_want(self, want):
+        new_want = remove_empties(want)
```
kerry-meyer, line 439: This "remove_empties" invocation is not needed because it was already done on line 143 in "set_config" when obtaining the "want" that is passed to "set_state" which, in turn, passes it to this function.

kerry-meyer (changes requested):
This change set is almost ready for merging. The current code looks fine, with the three remaining issues as the only exceptions to this. All test results look good.

I am posting only two minor comments on the current version of the change set. I also think it is worth evaluating the changes needed to use openconfig REST APIs for the "config" file. (The needed changes appear to be small and simple.)

plugins/module_utils/network/sonic/config/mirroring/mirroring.py
kerry-meyer: This looks like it wouldn't require much code change, and the "type" option is not really needed. It will be less confusing to users and future maintainers if we use the same Yang basis for both files.

kerry-meyer (approved):
Thank you for making the additional changes.

The current proposed change set and corresponding test results all look good.

Thank you for completing the implementation of the sonic_mirroring resource module.

Approved.

PR #533, opened 2025-03-18
Snc 40219 no autostate functionality for SONiC vlans

awhaley-dell:
##### SUMMARY

Adding the "autostate" functionality to the sonic_vlans module.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

sonic_vlans

##### OUTPUT

```paste below
UNIT TEST RESULTS:

============================= test session starts ==============================
platform linux -- Python 3.12.2, pytest-8.1.1, pluggy-1.4.0
rootdir: /root/ansible_collections/dellemc/enterprise_sonic
configfile: ../../../ansible/test/lib/ansible_test/_data/pytest/config/default.ini
plugins: xdist-3.5.0, mock-3.14.0
created: 4/4 workers
4 workers [22 items]

......................                                                   [100%]
- generated xml file: /root/ansible_collections/dellemc/enterprise_sonic/tests/output/junit/python3.12-controller-units.xml -
============================== 22 passed in 3.05s ==============================

```
Regression report
Date: 2025-09-12 Time: 17-00-42
Summary report
Module	Total testcases	Passed	Failed
vlans	20	20	0
Testcase Complete Summary: Total: 20, Passed: 20, Failed: 0

Detailed report of vlans
Testcase summary: Total: 20, Passed: 20, Failed: 0
[autostate-final-passing-regression-results.pdf](https://github.com/user-attachments/files/22352537/autostate-final-passing-regression-results.pdf)

```
<!--- Measure the code coverage before and after the change by running the UT and ensure that the "coverage after the change" is not less than the coverage "before the change". Note that the unit testing coverage can be manually executed using the pytest tool or ansible-test tool. -->

##### Checklist:

- Passing results for the regression tests in the sonic_vlan module




- Passing results for the unit tests in the sonic_vlan module

### CODE REVIEW

File: plugins/module_utils/network/sonic/config/vlans/vlans.py [hunk 1]
```diff
@@ -300,20 +323,50 @@ def get_create_vlans_requests(self, configs):
         if not configs:
             return requests
         for vlan in configs:
-            vlan_id = vlan.get("vlan_id")
+            vlan_id = vlan.get("vlan_id", None)
```
kerry-meyer, line 326: Here and elsewhere: If the desired default "get" value is None, it isn't necessary to explicitly add the default since that is the Python default for a "get". (It is, however necessary to specify it if the desired default is, for example, an empty list or an empty string.)

```suggestion
            vlan_id = vlan.get("vlan_id")
```

tests/regression/roles/sonic_vlans/defaults/main.yml
```diff
@@ -38,54 +43,62 @@ tests:
...
     description: Replace VLANs
     state: replaced
     input:
-        - vlan_id: 500
-        - vlan_id: 502
-          description: "modified vlan502 description"
+      - vlan_id: 500
+        autostate: true
+      - vlan_id: 502
+        description: "modified vlan502 description"
```
kerry-meyer, line 57: It would be good to verify that a 'replace' play that modifies autostate from 'true' to 'false' works correctly.

```suggestion
      - vlan_id: 500
        autostate: true
      - vlan_id: 501
        autostate: false
      - vlan_id: 502
        description: "modified vlan502 description"
```

tests/regression/roles/sonic_vlans/defaults/main.yml
```diff
@@ -38,54 +43,62 @@ tests:
...
     input:
-        - vlan_id: 501
-        - vlan_id: 502
-          description: "overridden vlan502 description"
-        - vlan_id: 504
+      - vlan_id: 501
+      - vlan_id: 502
+        description: "overridden vlan502 description"
+        autostate: true
+      - vlan_id: 504
```
kerry-meyer, line 68: It would be good to verify that setting (chaning) 'autostate' to 'false' works correctly in 'overridden' state.

```suggestion
      - vlan_id: 501
      - vlan_id: 502
        description: "overridden vlan502 description"
        autostate: true
      - vlan_id: 503
        autostate: false
      - vlan_id: 504
```

kerry-meyer (changes requested):
I am posting some change requests and a few questions on the proposed code modifications and additions.

In addition, please fix the sanity and UT failures, and please post regression test results that verify the 'check' and 'diff' mode handling for this module. (Instructions for doing that are available on the "#sonic-ansible-q-and-a" Slack channel.

tests/regression/roles/sonic_vlans/defaults/main.yml
```diff
@@ -38,54 +43,62 @@ tests:
       - vlan_id: 500
         description: "modified vlan500 description"
       - vlan_id: 501
```
kerry-meyer, line 45: It would provide better coverage to add an explicit deletion case for 'autostate' == true. I realize that this should be a functional no-op, but some verification that the correct behavior occurs for this case would be good. For full vlan deletion, the case can be moved from vlan 501 (now set to 'autostate' = true, so needed in the new case I'm requesting here) to vlan 502.

```suggestion
      - vlan_id: 500
        description: "modified vlan500 description"
      - vlan_id: 501
        autostate: true
      - vlan_id: 502
```

tests/regression/roles/sonic_vlans/defaults/main.yml
```diff
@@ -38,54 +43,62 @@ tests:
...
+      - vlan_id: 503
+      # Odd case where delete sets things to default true
+        autostate: true
```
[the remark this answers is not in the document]
awhaley-dell: Your assumption of the behavior is incorrect. The "delete" REST api sets autostate to true. 
This is the behavior for a few endpoints where DELETE sets the value back to the default value.
It can be counter-intuitive so I made sure to explicitly mention that in this comment so if someone assumes the behavior is like you thought they would be notified of the endpoints actual behavior.

Here is a little flowchart to show what the actual DELETE behavior of this endpoint is for autostate:
"enable" -> DELETE -> "enable"
"disable" -> DELETE -> "enable"

plugins/module_utils/network/sonic/config/vlans/vlans.py
awhaley-dell: I hear you in that it is not necessary to specify None during a get() request for a dictionary. However, this file has multiple references like this so to keep in line with prior style I explicitly mentioned it here. I can remove them which for aesthetics  but that shouldn't change any functionality.

plugins/module_utils/network/sonic/facts/vlans/vlans.py
```diff
@@ -85,16 +87,19 @@ def render_config(self, spec, conf):
...
+            config['autostate'] = conf.get('autostate')
         except TypeError:
             config['vlan_id'] = None
             config['description'] = None
+            config['autostate'] = None
         return utils.remove_empties(config)
 
     def get_vlans(self):
         """Get all the l2_interfaces available in chassis"""
+        # Grab description data from the openconfig-interfaces endpoint
```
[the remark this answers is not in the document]
awhaley-dell: The comment is not a part of the function header, it is on the next line. I can add a newline to better delineate between the two sections for clarity.

This first data collection step that uses openconfig-interfaces is original to this module. I didn't modify it and built on top of this since I was just working on implementing autostate. I was keeping the scope of my changes very narrow.

I'll see how much work it would be to move the sub-options to the sonic-vlan URI. Re-working the from openconfig-interfaces to sonic-vlans would be ideal but that would require reworking the config and unit tests files at the minimum.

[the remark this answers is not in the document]
kerry-meyer: I understand that the comment in question isn't a function header, but the real point about the comment is that, as one starts to read the function, it is distracting to see a comment highlighting the fact that two different endpoints are being used. This is less important than the structural complexity of merging the two results.

It wouldn't be very much work to revise the facts file to get all information from the same base URL and it would simplify the code. The corresponding UT changes don't look very big, either. This new addition is the "special case" that should be the trigger for consolidating this handling to keep the code readable and maintainable for the future. I am not requesting a similar change in the "config" file because it won't make a significant difference or simplification in the code structure. The code structure in this "facts" file, however, has become undesirably and unnecessarily complicated by using both URLs.

plugins/module_utils/network/sonic/config/vlans/vlans.py
kerry-meyer: Okay. I will drop this Issue. It is common enough in Python to do this, even though it is unnecessary. I agree that there is no functional impact and understand that this was the precedent set in this file, so it can stay "as is".

plugins/module_utils/network/sonic/config/vlans/vlans.py
```diff
@@ -75,7 +75,7 @@ def get_vlans_facts(self):
         :returns: The current configuration as a dictionary
         """
         facts, _warnings = Facts(self._module).get_facts(self.gather_subset, self.gather_network_resources)
-        vlans_facts = facts['ansible_network_resources'].get('vlans')
+        vlans_facts = facts['ansible_network_resources'].get('vlans', None)
```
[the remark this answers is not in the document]
kerry-meyer: As mentioned below, I will drop this Issue. It is common enough in Python to do this, even though it is unnecessary. I agree that there is no functional impact and understand that this was the precedent set in this module, so it can stay "as is".

tests/regression/roles/sonic_vlans/defaults/main.yml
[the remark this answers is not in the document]
kerry-meyer: My assumption for the behavior matches the actual behavior.

To delete a (non-default) 'autostate' value of 'false', the playbook needs to specify:

autostate: false

That is what should be done to set it to the default state of "true".

If that doesn't work, something is wrong. In this case, it is clear what is wrong and I have posted other comments to address the cause of the problem.

awhaley-dell:
The failing checkers are not related to my code it seems. They seem related to the ansible infra in general:

Installing collected packages: tomlkit, PyYAML, platformdirs, mccabe, isort, dill, astroid, pylint
Successfully installed PyYAML-6.0.2 astroid-3.3.11 dill-0.4.0 isort-6.0.1 mccabe-0.7.0 platformdirs-4.4.0 pylint-3.3.8 tomlkit-0.13.3
ERROR: Found 2 pylint issue(s) which need to be resolved:
ERROR: plugins/module_utils/network/sonic/utils/utils.py:18:0: ansible-bad-import-from: Import iteritems from the Python standard library equivalent instead of ansible.module_utils.six
ERROR: tests/unit/modules/conftest.py:11:0: ansible-bad-import-from: Import string_types from the Python standard library equivalent instead of ansible.module_utils.six
See documentation for help: https://docs.ansible.com/ansible-core/devel/dev_guide/testing/sanity/pylint.html

plugins/module_utils/network/sonic/facts/vlans/vlans.py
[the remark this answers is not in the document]
kerry-meyer: Consolidation of the facts gathering to a single endpoint and the accompanying revisions have fully addressed the concern expressed for this issue.

Resolved.

plugins/module_utils/network/sonic/config/vlans/vlans.py
```diff
@@ -168,25 +169,28 @@ def _state_replaced(self, want, have, diff):
...
+
+            if replaced_vlans:
+                del_requests = self.get_delete_vlans_requests(replaced_vlans, delete_vlan=True)
+                requests.extend(del_requests)
+                commands.extend(update_states(replaced_config, "deleted"))
             requests.extend(del_requests)
             commands.extend(update_states(replaced_config, "deleted"))
 
+        diff = get_diff(want, have, TEST_KEYS)
+        diff = self.deal_with_default_entries(diff, method="remove")
```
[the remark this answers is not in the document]
kerry-meyer: Although the instruction in question has been modified, the changed version still invokes the "deal_with_default_entries" method in a way that is questionable. It should be unnecessary and is indicative of an underlying bug.

I will post a separate comment for the revised instruction, but am leaving this "Issue" open because it hasn't been adequately addressed.

plugins/module_utils/network/sonic/config/vlans/vlans.py
```diff
@@ -263,57 +277,100 @@ def _state_deleted(self, want, have, diff):
...
+                    request = {"path": path,
+                               "method": method,
+                               }
+                    requests.append(request)
+
+                # Delete sets autostate to "disable", not the same behavior as REST DELETE command at that endpoint but makes sense for boolean.
+                if vlan.get("autostate", None) is not None:
+                    path = "data/sonic-vlan:sonic-vlan/VLAN/VLAN_LIST=Vlan{}/{}".format(vlan_id, "autostate")
+                    payload = {"sonic-vlan:autostate": "disable"}
+                    request = {"path": path, "method": "PATCH", "data": payload}
```
kerry-meyer, line 311: This logic is incorrect.

The specified and correct handling for a "delete" does not depend on what "makes sense for boolean". It depends on the situation.

Deletion of an option should occur if (and only if) the deletion request ("mention" of the option) from the "deleted" state play specifies a value for the option that matches the current configured value.

If the option is currently configured to a non-default value, "deletion" causes it to be truly deleted (if the target device supports that) or else causes it to be set to the default value, which in this case is "True".

A request to "delete" an option that is already deleted or that is set to the default value (if true deletion is not supported) has no effect and should not generate a request. I think that the code which invokes this function already ensures that the command will not be sent to this function for the "no-op" case described here. (If it doesn't ensure that, it should be modified accordingly.)

The current implementation here does the opposite of what should be done for a value deletion. Instead of "patching" a non-default value, it should either "delete" the current value or patch it to the default value if it is in the non-default state (false for the Ansible argspec or "disable" for the REST state) and the executing play specifies deletion of "false". If the requested deletion doesn't match the current state or the current configured state is the default state (which can not be "deleted"), this function should avoid enqueuing a request.

plugins/module_utils/network/sonic/config/vlans/vlans.py
```diff
@@ -168,25 +169,28 @@ def _state_replaced(self, want, have, diff):
         """
         commands = []
         requests = []
-
+        # Add default values to want so it'll be properly comparable to have and re-generate diff.
+        diff = get_diff(self.deal_with_default_entries(want, method="add"), have, TEST_KEYS)
```
kerry-meyer, line 173: The "deal_with_default_entries" invocation here seems incorrect.

If a "replaced" play doesn't specify some option and that option is not currently configured (to a non-default value, in this case), no action should occur. If it doesn't specify some option and that option is currently configured to a non-default value, it should cause it to be logically "deleted" (set to the default value, in this case), so the second pass of the "replaced" handling (the "merge" after "delete" pass) shouldn't need to take any further action. Either way, for the diff, having the option absent from the input playbook should not result in a "forward" (want vs have) diff entry. (That should only happen if the new requested value is specified and differs from the current configuration.)

kerry-meyer:
You are correct that the sanity errors have nothing to do with this change set.

We are looking into a solution for that problem, and it will not be a factor in determining when this PR can be approved.

tests/regression/roles/sonic_vlans/defaults/main.yml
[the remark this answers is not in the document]
awhaley-dell: I see, you did understand the behavior. My apologies. I'll take a look at the other review comments to address the root cause.

plugins/module_utils/network/sonic/config/vlans/vlans.py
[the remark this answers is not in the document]
awhaley-dell: There was a bug where this default value was non-existent in REST until modified once: https://jira.cec.lab.emc.com/browse/SNC-40907
It has been addressed and this behavior needs to be re-evaluated.

tests/regression/roles/sonic_vlans/defaults/main.yml
kerry-meyer: The requested case still isn't present. This could be accomplished by setting autostate to true for vlan 504 in test_case_02 and setting it to false in test_case_04.

plugins/module_utils/network/sonic/config/vlans/vlans.py
```diff
@@ -200,33 +204,31 @@ def _state_overridden(self, want, have, diff):
...
-        if del_descr_vlans:
-            del_requests = self.get_delete_vlans_requests(del_descr_vlans, False)
+        del_commands = []
+        # Find VLANS to delete or modify using reverse_diff and two lists to separate complete deletion and modification
+        for config in reverse_diff:
+            want_vlan_obj = search_obj_in_list(config['vlan_id'], want, 'vlan_id')
+            if want_vlan_obj:
+                if want_vlan_obj.get('description', None) is None and config.get('description', None) is not None:
+                    del_commands.append({"vlan_id": config.get("vlan_id"), "description": config.get("description")})
+                if want_vlan_obj.get('autostate', None) is True and config.get('autostate', None) is False:
```
kerry-meyer, line 219: autostate deletion should also occur here If the reverse_diff vlan has no autostate option, but autostate is configured (to False).

```suggestion
                if ((want_vlan_obj.get('autostate', None) is True or
                     want_vlan_obj.get('autostate', None) is None) and
                    config.get('autostate', None) is False):
```

kerry-meyer (changes requested):
I have resolved most of the previously existing comments. I am posting a few new comments, all of them for minor changes.

PR #578, opened 2025-08-06
Fix EVPN ESI multihome delete all bug

stalabi1:
##### SUMMARY

I updated the URL used for delete all. Prior to this bugfix, the delete all URL was 'data/openconfig-network-instance:network-instances/network-instance=default/evpn/evpn-mh/config'. Delete all with this URL did appear to delete all the EVPN multihome config (the get output returned {}) however, some key entries still remained in the DB. This issue was uncovered by the sonic_mclag regression test which is executed after the sonic_evpn_esi_multihome regression test. MCLAG cannot be configured if there's evpn config on the device.

Upon further investigation, I saw that the  CLI deletes the EVPN multihome config by specifying this special URL:
'data/openconfig-network-instance:network-instances/network-instance=default/evpn/evpn-mh/config?deleteEmptyEntry=True' which will also delete the necessary key entries in the DB. Instead of making use of this special URL, I opted to perform delete all at the next level up, 'data/openconfig-network-instance:network-instances/network-instance=default/evpn/evpn-mh' since it seems more standard. Delete with this URL also removes the necessary key entries from the DB.

I also corrected the numbering of the test cases for regression and removed a redundant replaced test case in the UT.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

sonic_evpn_esi_multihome

##### OUTPUT

Failing report prior to bugfix:
[regression-2025-08-06-12-56-24.html.pdf](https://github.com/user-attachments/files/21640975/regression-2025-08-06-12-56-24.html.pdf)

Passing report after bugfix:
[regression-2025-08-06-14-35-52.html.pdf](https://github.com/user-attachments/files/21640981/regression-2025-08-06-14-35-52.html.pdf)

##### Checklist:

- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have maintained backward compatibility or have provided any relevant "breaking_changes" descriptions in a "fragment" file in the "changelogs/fragments" directory of this repository.

### DIFF
--- a/tests/regression/roles/sonic_evpn_esi_multihome/defaults/main.yml
+++ b/tests/regression/roles/sonic_evpn_esi_multihome/defaults/main.yml
@@ -52,6 +52,6 @@ tests:
       startup_delay: 500
 
   - name: test_case_07
-    description: Delete EVPN esi multihome configuration
+    description: Delete all EVPN esi multihome configuration
     state: deleted
     input: {}


kerry-meyer (approved):
The proposed code changes and corresponding test results look good.

Thank you for fixing this problem.

Approved.

PR #630, opened 2026-03-31
630-fix-ospfv2-config-for-protocol-name.yaml

Mansi062001:
Added logic to support redistribution configuration for multiple routing protocols, ensuring that redistribution is applied even when no additional protocol‑specific parameters are provided.

Here is the Full regression log:
[ansible_regression.txt](https://github.com/user-attachments/files/26370036/ansible_regression.txt)

UT log:
[ansible_log2.txt](https://github.com/user-attachments/files/26370053/ansible_log2.txt)


stalabi1 (changes requested):
Please resolve sanity errors.

stalabi1 (approved):

